PR #33, opened 2020-10-21
feat: added depositAll

Macarse:


contracts/Vault.vy
```diff
@@ -302,11 +302,15 @@ def _issueSharesForAmount(_to: address, _amount: uint256) -> uint256:
...
 
-@external
-def deposit(_amount: uint256) -> uint256:
+@internal
+def _deposit(_sender: address, _amount: uint256) -> uint256:
     assert not self.emergencyShutdown  # Deposits are locked out
     assert self._totalAssets() + _amount <= self.depositLimit  # Max deposit reached
 
+    # Depositer should have funds
+    assert self.token.balanceOf(_sender) > 0
```
fubuloubu, line 311: Cheaper, no external call required. Invariant that balance if greater than 0 is already demonstrated below by success of `transferFrom` (asserting that `_amount > 0` here ensures that we are always transfering >0 balance below)
```suggestion
    assert _amount > 0
```

contracts/Vault.vy
```diff
@@ -302,11 +302,15 @@ def _issueSharesForAmount(_to: address, _amount: uint256) -> uint256:
...
+    assert self.token.balanceOf(_sender) >= _amount
```
fubuloubu, line 312: This is inherently true if `self.token.transferFrom(_sender, self, _amount)` succeeds (so, not necessary)

fubuloubu (approved):
Minor changes, looks good.

Should add `withdrawAll()` in a separate PR?

contracts/Vault.vy
```diff
@@ -302,11 +302,15 @@ def _issueSharesForAmount(_to: address, _amount: uint256) -> uint256:
...
 
 
-@external
-def deposit(_amount: uint256) -> uint256:
+@internal
+def _deposit(_sender: address, _amount: uint256) -> uint256:
     assert not self.emergencyShutdown  # Deposits are locked out
     assert self._totalAssets() + _amount <= self.depositLimit  # Max deposit reached
 
+    # Depositer should have funds
```
fubuloubu, line 310:
```suggestion
    # Ensure we are depositing something
```

Macarse: fixed

PR #36, opened 2020-10-22
feat: withdrawAll

Macarse:


Macarse:
not sure why this is failing in ci/cd
```
46 passed in 171.13s (0:02:51)
```

fubuloubu:
Failure here is really strange...

banteg:
Imo redundant method.

fubuloubu:
@banteg we agree that it may be a little redundant, but it is useful for some as a convienence. We were looking do actually abstract #38 a bit more into a signature like `deposit(_to: address = msg.sender, _amount: uint256 = MAX_UINT256)`, where `_amount == MAX_UINT256` triggers depositing everything. That would reduce everything by a lot.

fubuloubu:
closing in favor of #38 

thanks for giving it a shot!

if you want to try and give https://github.com/iearn-finance/yearn-vaults/pull/36#issuecomment-716081402 a shot in another PR, definitely willing to help on that.

PR #41, opened 2020-10-26
feat: generalise deposit and withdraw methods

chriswessels:
Building off of #38 and the comment by @fubuloubu https://github.com/iearn-finance/yearn-vaults/pull/36#issuecomment-716081402, this PR:

- [x] Removes `depositAll` and `withdrawAll`
- [x] Removes indirection and makes `deposit` and `withdraw` `@external` again
- [x] Generalises argument signature to `(_amount: uint256 = MAX_UINT256, _recipient: address = msg.sender)` for `deposit` and `withdraw`. When `_amount == MAX_UINT256`, the maximum token or share balance is used.
- [x] Updates all tests with new deposit/withdraw usage


contracts/Vault.vy
```diff
@@ -314,13 +314,19 @@ def _issueSharesForAmount(_to: address, _amount: uint256) -> uint256:
     return shares
 
 
-@internal
-def _depositFrom(_sender: address, _recipient: address, _amount: uint256) -> uint256:
+@external
+def deposit(_recipient: address = msg.sender, _amount: uint256 = MAX_UINT256) -> uint256:
```
Macarse, line 318: nice! I was playing with this yesterday and I didn't understand why this signature doesn't allow:
```
deposit(_amount=1234)
```

Somehow vyper is not generating a method for `deposit(_amount)` which would be when `_recipient` is the default.

tests/functional/vault/test_misc.py
```diff
@@ -48,8 +48,8 @@ def test_reject_ether(gov, vault):
                 1,
             ],
         ),
-        ("deposit", [1]),
-        ("withdraw", [1]),
+        ("deposit", []),
+        ("withdraw", []),
```
Macarse, line 52: for test completeness, we should have all the different possibilities

fubuloubu: Great point!

contracts/Vault.vy
fubuloubu: ah yes, I see your problem now @Macarse, web3py automagically finds the correct signature based on the type of all the arguments (throwing if there's multiple matches). You don't use kwargs directly for this (e.g. `_amount=12354` won't work)

tests/functional/vault/test_misc.py
chriswessels: Will add additional cases!

contracts/Vault.vy
fubuloubu: @chriswessels swap the order of these arguments so that `deposit(1234)` can be called to deposit the amount mentioned, it is definitely expected to call with this more often than with delegated amounts

chriswessels: @fubuloubu Done!

fubuloubu (approved):

PR #42, opened 2020-10-26
Updates to Strategy API

fubuloubu:
- [x] Changed `Vault.version()` and `Strategy.version()` to `*.apiVersion()` to clarify intent of this function (mostly for Strategy usage)
- [x] Added `Strategy.name()` required override function so that users give their strategies a custom name
- [x] Fixed `Strategy.protectedTokens()` optional override function so that users can define their own list (or use the default)


dmolina79 (approved):
LGTM!

chriswessels (approved):
lgtm 👍

PR #44, opened 2020-10-27
refactor: remove additional check on approvals

fubuloubu:
Removes the overly onerous approval check as a mitigation for a minor vulnerability in ERC20 implementation.

### DIFF
--- a/tests/functional/vault/test_shares.py
+++ b/tests/functional/vault/test_shares.py
@@ -187,12 +187,6 @@ def test_transferFrom(accounts, token, vault, fn_isolation):
     with brownie.reverts():
         vault.transferFrom(a, b, vault.balanceOf(a) // 2, {"from": c})
 
-    # Show that approval ordering protection works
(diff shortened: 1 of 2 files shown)


tests/functional/vault/test_shares.py
```diff
@@ -187,12 +187,6 @@ def test_transferFrom(accounts, token, vault, fn_isolation):
     with brownie.reverts():
         vault.transferFrom(a, b, vault.balanceOf(a) // 2, {"from": c})
 
-    # Show that approval ordering protection works
```
Macarse, line 190: if the change was redundant, why did you need to remove this part of the test?

fubuloubu: wrong choice of words, I meant overly onerous

lbertenasco (approved):
This is fine, the "attack" es not an issue.
It does NOT make any sense to do this kind of checks on our contracts (where users usually approve an infinite allowance).
Also where we mostly interact between "internal" contracts by only approving the exact amount that will be utilized making the allowance 0 again.

PR #50, opened 2020-10-30
Refactor Strategy API

fubuloubu:
A series of fixes to simplify how strategist use `BaseStrategy`
- [x] Simplify protected tokens interface
- [x] Remove `expectedReturn`
- [x] Fix `StrategyParams.lastReport` in `BaseStrategy`
- [x] Remove `lifetimeEarned` from `BaseStrategy.Harvest` event (can query this using `Vault.StrategyReported` instead)
- [x] Move `*Trigger` functions to `BaseStrategy` template (overrideable)
- [x] Reconfigure `BaseStrategy` API touchpoints to not have to read/write storage values directly
- [x] Add a method to force `harvest` to trigger based on time

Bumped API Version to 0.1.3

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -317,16 +317,14 @@ abstract contract BaseStrategy {
         outstanding = vault.report(want.balanceOf(address(this)).sub(reserve));
     }
 
-    // Override this to add all tokens this contract manages on a *persistant* basis
-    // (e.g. not just for swapping back to want ephemerally)
-    // NOTE: Must inclide `want` token
-    function protectedTokens() internal virtual view returns (address[] memory) {
-        address[] memory protected = new address[](1);
-        protected[0] = address(want);
-        return protected;
-    }
+    // Override this to add all tokens/tokenized positions this contract manages
+    // on a *persistant* basis (e.g. not just for swapping back to want ephemerally)
+    // NOTE: Do *not* include `want`, already included in `sweep` below
+    function protectedTokens() internal virtual view returns (address[] memory);
 
     function sweep(address _token) external {
+        require(_token != address(want), "!want");
+
         address[] memory _protectedTokens = protectedTokens();
         for (uint256 i; i < _protectedTokens.length; i++) require(_token != _protectedTokens[i], "!protected");
 
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -10,7 +10,7 @@ struct StrategyParams {
     uint256 activation;
     uint256 debtLimit;
     uint256 rateLimit;
-    uint256 lastSync;
+    uint256 lastReport;
     uint256 totalDebt;
     uint256 totalReturns;
 }
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -133,18 +133,17 @@ abstract contract BaseStrategy {
     // So indexers can keep track of this
     event Harvested(uint256 profit);
 
-    // Adjust this to keep some of the position in reserve in the strategy,
+    // Adjust this using `setReserve(...)` to keep some of the position in reserve in the strategy,
     // to accomodate larger variations needed to sustain the strategy's core positon(s)
-    uint256 public reserve = 0;
+    uint256 private reserve = 0;
 
-    // This gets adjusted every time the Strategy reports to the Vault,
-    // and should be used during adjustment of the strategy's positions to "deleverage"
-    // in order to pay back the amount the next time it reports.
-    //
-    // NOTE: Do not edit this variable, for safe usage (only read from it)
-    // NOTE: Strategy should not expect to increase it's working capital until this value
-    //       is zero.
-    uint256 public outstanding = 0;
+    function getReserve() internal view returns (uint256) {
+        return reserve;
+    }
+
+    function setReserve(uint256 _reserve) internal {
+        if (_reserve != reserve) reserve = _reserve;
+    }
 
     bool public emergencyExit;
 
@@ -157,12 +156,12 @@ abstract contract BaseStrategy {
     }
 
     function setStrategist(address _strategist) external {
-        require(msg.sender == strategist || msg.sender == governance(), "!governance");
+        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
         strategist = _strategist;
     }
 
     function setKeeper(address _keeper) external {
-        require(msg.sender == strategist || msg.sender == governance(), "!governance");
+        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
         keeper = _keeper;
     }
 
@@ -187,10 +186,10 @@ abstract contract BaseStrategy {
      *       conditions (e.g. this function is possible to influence through flashloan
      *       attacks, oracle manipulations, or other DeFi attack mechanisms).
      *
-     * NOTE: It is up to governance to use this function in order to correctly order
-     *       this strategy relative to its peers in order to minimize losses for the
-     *       Vault based on sudden withdrawals. This value should be higher than the
-     *       total debt of the strategy and higher than it's expected value to be "safe".
+     * NOTE: It is up to governance to use this function to correctly order this strategy
+     *       relative to its peers in the withdrawal queue to minimize losses for the Vault
+     *       based on sudden withdrawals. This value should be higher than the total debt of
+     *       the strategy and higher than it's expected value to be "safe".
      */
     function estimatedTotalAssets() public virtual view returns (uint256);
 
@@ -204,7 +203,7 @@ abstract contract BaseStrategy {
      * strategy and reduce it's overall position if lower than expected returns
      * are sustained for long periods of time.
      */
-    function prepareReturn() internal virtual;
+    function prepareReturn(uint256 _debtOustanding) internal virtual returns (uint256 _profit);
 
     /*
      * Perform any adjustments to the core position(s) of this strategy given
@@ -213,7 +212,7 @@ abstract contract BaseStrategy {
      * was made is available for reinvestment. Also note that this number could
      * be 0, and you should handle that scenario accordingly.
      */
-    function adjustPosition() internal virtual;
+    function adjustPosition(uint256 _debtOutstanding) internal virtual;
 
     /*
      * Make as much capital as possible "free" for the Vault to take. Some slippage
@@ -243,9 +242,12 @@ abstract contract BaseStrategy {
     }
 
     function tend() external {
-        if (keeper != address(0)) require(msg.sender == keeper || msg.sender == strategist || msg.sender == governance());
-        // NOTE: Don't take profits with this call, but adjust for better gains
-        adjustPosition();
+        if (keeper != address(0)) {
+            require(msg.sender == keeper || msg.sender == strategist || msg.sender == governance(), "!authorized");
+        }
+
+        // Don't take profits with this call, but adjust for better gains
+        adjustPosition(vault.debtOutstanding());
     }
 
     /*
@@ -285,37 +287,44 @@ abstract contract BaseStrategy {
     }
 
     function harvest() external {
-        if (keeper != address(0)) require(msg.sender == keeper || msg.sender == strategist || msg.sender == governance());
+        if (keeper != address(0)) {
+            require(msg.sender == keeper || msg.sender == strategist || msg.sender == governance(), "!authorized");
+        }
 
+        uint256 profit = 0;
         if (emergencyExit) {
             exitPosition(); // Free up as much capital as possible
             // NOTE: Don't take performance fee in this scenario
         } else {
-            prepareReturn(); // Free up returns for Vault to pull
+            profit = prepareReturn(vault.debtOutstanding()); // Free up returns for Vault to pull
         }
 
         if (reserve > want.balanceOf(address(this))) reserve = want.balanceOf(address(this));
 
         // Allow Vault to take up to the "harvested" balance of this contract, which is
         // the amount it has earned since the last time it reported to the Vault
-        uint256 wantEarned = want.balanceOf(address(this)).sub(reserve);
-        outstanding = vault.report(wantEarned);
+        uint256 outstanding = vault.report(want.balanceOf(address(this)).sub(reserve));
 
-        adjustPosition(); // Check if free returns are left, and re-invest them
+        // Check if free returns are left, and re-invest them
+        adjustPosition(outstanding);
 
         emit Harvested(profit);
     }
 
     /*
      * Liquidate as many assets as possible to `want`, irregardless of slippage,
-     * up to `_amount`. Any excess should be re-invested here as well.
+     * up to `_amountNeeded`. Any excess should be re-invested here as well.
      */
-    function liquidatePosition(uint256 _amount) internal virtual;
+    function liquidatePosition(uint256 _amountNeeded) internal virtual returns (uint256 _amountFreed);
 
-    function withdraw(uint256 _amount) external {
+    function withdraw(uint256 _amountNeeded) external {
         require(msg.sender == address(vault), "!vault");
-        liquidatePosition(_amount); // Liquidates as much as possible to `want`, up to `_amount`
-        want.transfer(msg.sender, want.balanceOf(address(this)).sub(reserve));
+        // Liquidate as much as possible to `want`, up to `_amount`
+        uint256 amountFreed = liquidatePosition(_amountNeeded);
+        // Send it directly back (NOTE: Using `msg.sender` saves some gas here)
+        want.transfer(msg.sender, amountFreed);
+        // Adjust reserve to what we have after the freed amount is sent to the Vault
+        reserve = want.balanceOf(address(this));
     }
 
     /*
@@ -328,6 +337,7 @@ abstract contract BaseStrategy {
         require(msg.sender == address(vault) || msg.sender == governance());
         require(BaseStrategy(_newStrategy).vault() == vault);
         prepareMigration(_newStrategy);
+        want.transfer(_newStrategy, want.balanceOf(address(this)));
     }
 
     function setEmergencyExit() external {
@@ -336,7 +346,6 @@ abstract contract BaseStrategy {
         exitPosition();
         vault.revokeStrategy();
         if (reserve > want.balanceOf(address(this))) reserve = want.balanceOf(address(this));
-        outstanding = vault.report(want.balanceOf(address(this)).sub(reserve));
     }
 
     // Override this to add all tokens/tokenized positions this contract manages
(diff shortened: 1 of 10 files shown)


contracts/BaseStrategy.sol
```diff
@@ -317,16 +310,14 @@ abstract contract BaseStrategy {
...
-    // NOTE: Must inclide `want` token
-    function protectedTokens() internal virtual view returns (address[] memory) {
-        address[] memory protected = new address[](1);
-        protected[0] = address(want);
-        return protected;
-    }
+    // Override this to add all tokens/tokenized positions this contract manages
+    // on a *persistant* basis (e.g. not just for swapping back to want ephemerally)
+    // NOTE: Do *not* include `want`, already included in `sweep` below
+    function protectedTokens() internal virtual view returns (address[] memory);
```
Macarse, line 316: can you move it to the top with the other methods you need to override?

contracts/test/TestStrategy.sol
```diff
@@ -87,4 +83,8 @@ contract TestStrategy is BaseStrategy {
     function prepareMigration(address _newStrategy) internal override {
         want.transfer(_newStrategy, want.balanceOf(address(this)));
     }
+
+    function protectedTokens() internal override view returns (address[] memory) {
```
Macarse, line 87: why didn't you copy the exact method and body from the previous implementation?
Most strategists are going to copy and paste code from `TestStrategy` I would leave here the complete method.

Macarse (approved):

contracts/BaseStrategy.sol
fubuloubu: :+1:

fubuloubu: well, actually I had been placing them near where they were being used, so that it's easier to see the context

contracts/test/TestStrategy.sol
fubuloubu: I removed `want` from there and made it automatically checked for

contracts/BaseStrategy.sol
```diff
@@ -317,16 +310,14 @@ abstract contract BaseStrategy {
...
-    // (e.g. not just for swapping back to want ephemerally)
-    // NOTE: Must inclide `want` token
-    function protectedTokens() internal virtual view returns (address[] memory) {
-        address[] memory protected = new address[](1);
-        protected[0] = address(want);
-        return protected;
-    }
+    // Override this to add all tokens/tokenized positions this contract manages
+    // on a *persistant* basis (e.g. not just for swapping back to want ephemerally)
+    // NOTE: Do *not* include `want`, already included in `sweep` below
```
Grandthrax, line 374: leave the old code as a comment.

contracts/BaseStrategy.sol
```diff
@@ -317,16 +310,14 @@ abstract contract BaseStrategy {
...
+    function protectedTokens() internal virtual view returns (address[] memory);
 
     function sweep(address _token) external {
```
Grandthrax, line 377: I don't think there is any need for this to be open to public. I can only think of a few very fringe situations where governance wouldn't want to do it themselves but someone else would. And there is a good chance someone will mess up the protectedTokens when writing a strategy at some point in the future.

contracts/BaseStrategy.sol
```diff
@@ -258,9 +256,33 @@ abstract contract BaseStrategy {
...
+        StrategyParams memory params = vault.strategies(address(this));
+
+        // Should not trigger if strategy is not activated
+        if (params.activation == 0) return false;
+
+        // If some amount is owed, pay it back
+        // NOTE: Since debt is adjusted in step-wise fashion, it is appropiate to always trigger here,
+        //       because the resulting change should be large (might not always be the case)
+        uint256 outstanding = vault.debtOutstanding();
+        if (outstanding > 0) return true;
```
Grandthrax, line 296: This needs some more thought. If a strategy is slowly withdrawing - for instance in a timelock situation. we don't want harvest being called every block.

contracts/BaseStrategy.sol
```diff
@@ -258,9 +256,33 @@ abstract contract BaseStrategy {
...
+        // Check for profits and losses
+        uint256 total = estimatedTotalAssets();
+        if (total < params.totalDebt) return true; // We have a loss to report!
+
+        uint256 profit = 0;
+        if (total > params.totalDebt) profit = total.sub(params.totalDebt); // We've earned a profit!
+
+        // Otherwise, only trigger if it "makes sense" economically (gas cost is <N% of value moved)
+        uint256 credit = vault.creditAvailable();
+        return (100 * callCost < credit.add(profit));
```
Grandthrax: 100 should probably be a variable. I think it should be added to base strategy. Gas factor for instance. 

100 will often not be high enough. It works out as gas costs will >= 1% of returns

contracts/BaseStrategy.sol
fubuloubu: This will never happen, because (and the comment reflects this) debt changes occur in a step-wise fashion. So basically if you have 100k and then your debt limit drops to 50k, you will have a debt outstanding of 50k and should try to unwind enough to pay that back

contracts/BaseStrategy.sol
fubuloubu: See b9d172c

contracts/BaseStrategy.sol
fubuloubu: See 5664f91

contracts/BaseStrategy.sol
Grandthrax: Debt is not always adjusted stepwise. For instance in an emergency shutdown situation. But even if it is strategy may not be able to free up 50k. But I suppose those edge cases can rewrite the whole method. Might be worth leaving virtual in the function declaration, that gives strategist the option to override if they have an edge case. Like you've done with tend trigger.

contracts/BaseStrategy.sol
fubuloubu: Solved in 19e0ca5

fubuloubu:
NOTE: No critical updates to Vault

contracts/BaseStrategy.sol
```diff
@@ -258,42 +276,76 @@ abstract contract BaseStrategy {
      * would be negatively affected if `harvest()` is not called shortly, then this can return `true`
      * even if the keeper might be "at a loss" (keepers are always reimbursed by yEarn)
      *
+     * NOTE: `callCost` must be priced in terms of `want`
+     *
      * NOTE: this call and `tendTrigger` should never return `true` at the same time.
      */
-    function harvestTrigger(uint256 gasCost) public virtual view returns (bool);
+    function harvestTrigger(uint256 callCost) public view returns (bool) {
```
Grandthrax, line 283: Not sure if you saw my previous comment and I'm not great at github so adding it here. 
_"Debt is not always adjusted stepwise. For instance in an emergency shutdown situation. But even if it is strategy may not be able to free up 50k. But I suppose those edge cases can rewrite the whole method. Might be worth leaving virtual in the function declaration, that gives strategist the option to override if they have an edge case. Like you've done with tend trigger."_

For instance in the strategy I am writing at the moment I would want to overwrite to something like:
if (outstanding > 0 && crETH.getCash() > 0) return true;

To cover liquidity crisis situation on lending platform

contracts/BaseStrategy.sol
```diff
@@ -131,20 +131,27 @@ abstract contract BaseStrategy {
     IERC20 public want;
 
     // So indexers can keep track of this
-    event Harvested(uint256 wantEarned, uint256 lifetimeEarned);
+    event Harvested(uint256 profit);
+
+    // The maximum number of blocks between harvest calls
+    // NOTE: Override this value with your own, or set dynamically below
+    uint256 public maxReportDelay = 6300; // ~ once a day
```
banteg, line 138:
```suggestion
    // The minimum number of blocks between harvest calls
    // NOTE: Override this value with your own, or set dynamically below
    uint256 public minReportDelay = 6300; // ~ once a day
```

contracts/BaseStrategy.sol
```diff
@@ -157,15 +164,25 @@ abstract contract BaseStrategy {
...
 
     function setKeeper(address _keeper) external {
-        require(msg.sender == strategist || msg.sender == governance(), "!governance");
+        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
         keeper = _keeper;
     }
 
+    function setMaxReportDelay(uint256 _delay) external {
+        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
+        maxReportDelay = _delay;
```
banteg, line 178:
```suggestion
    function setMinReportDelay(uint256 _delay) external {
        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
        minReportDelay = _delay;
```

contracts/BaseStrategy.sol
```diff
@@ -258,42 +276,76 @@ abstract contract BaseStrategy {
...
      */
-    function harvestTrigger(uint256 gasCost) public virtual view returns (bool);
+    function harvestTrigger(uint256 callCost) public view returns (bool) {
+        StrategyParams memory params = vault.strategies(address(this));
+
+        // Should not trigger if strategy is not activated
+        if (params.activation == 0) return false;
+
+        // Should trigger if hadn't been called in a while
+        if (block.number.sub(params.lastReport) >= maxReportDelay) return true;
```
banteg, line 290:
```suggestion
        if (block.number.sub(params.lastReport) >= minReportDelay) return true;
```

tests/functional/strategy/test_config.py
```diff
@@ -70,3 +39,35 @@ def test_strategy_setEmergencyExit(strategy, gov, strategist, rando, chain):
...
     strategy.setEmergencyExit({"from": strategist})
     assert strategy.emergencyExit()
+
+
+@pytest.mark.parametrize(
+    "getter,setter,val",
+    [
+        ("strategist", "setStrategist", None),
+        ("keeper", "setKeeper", None),
+        ("maxReportDelay", "setMaxReportDelay", 1000),
```
banteg, line 49:
```suggestion
        ("minReportDelay", "setMinReportDelay", 1000),
```

contracts/BaseStrategy.sol
```diff
@@ -174,13 +191,6 @@ abstract contract BaseStrategy {
         return vault.governance();
     }
 
-    /*
-     * Provide an accurate expected value for the return this strategy
-     * would provide to the Vault the next time `report()` is called
-     * (since the last time it was called)
-     */
-    function expectedReturn() public virtual view returns (uint256);
```
banteg, line 182: What's the rationale of removing `expectedReturn`?

contracts/BaseStrategy.sol
```diff
@@ -239,15 +249,23 @@ abstract contract BaseStrategy {
      * would be negatively affected if `tend()` is not called shortly, then this can return `true`
      * even if the keeper might be "at a loss" (keepers are always reimbursed by yEarn)
      *
+     * NOTE: `callCost` must be priced in terms of `want`
```
banteg, line 252: Should a strategy implement a function that quotes its `want` or should a keeper bot do this?

contracts/test/TestStrategy.sol
```diff
@@ -24,67 +24,51 @@ contract TestStrategy is BaseStrategy {
...
+        countdownTimer = countdownTimer.sub(1); // NOTE: This should never be called after it hits 0
     }
 
     function prepareMigration(address _newStrategy) internal override {
-        want.transfer(_newStrategy, want.balanceOf(address(this)));
+        // Nothing needed here because no additional tokens/tokenized positions for mock
+    }
+
+    function protectedTokens() internal override view returns (address[] memory) {
+        return new address[](0); // No additional tokens/tokenized positions for mock
```
banteg, line 72: I think an example should include at least one token
```suggestion
        address[] memory protected = new address[](1);
        protected[0] = 0x00000000000000000000000000000000DeaDBeef;
        return protected;
```

banteg (commented):
LGTM except few nitpicks

contracts/BaseStrategy.sol
```diff
@@ -258,42 +276,76 @@ abstract contract BaseStrategy {
...
+
+        // If some amount is owed, pay it back
+        // NOTE: Since debt is adjusted in step-wise fashion, it is appropiate to always trigger here,
+        //       because the resulting change should be large (might not always be the case)
+        uint256 outstanding = vault.debtOutstanding();
+        if (outstanding > 0) return true;
+
+        // Check for profits and losses
+        uint256 total = estimatedTotalAssets();
+        if (total < params.totalDebt) return true; // We have a loss to report!
```
Grandthrax, line 300: this triggers for me a lot because of rounding errors. there should be a threshold added

for instance. minting a ctoken gives can give you a slight on paper loss due to rounding errors on converting exchange rate of ctoken back to underlying asset

contracts/BaseStrategy.sol
```diff
@@ -131,20 +131,27 @@ abstract contract BaseStrategy {
     IERC20 public want;
 
     // So indexers can keep track of this
-    event Harvested(uint256 wantEarned, uint256 lifetimeEarned);
+    event Harvested(uint256 profit);
```
Macarse, line 134: mmm should we do the `losses` change in this version?

contracts/BaseStrategy.sol
```diff
@@ -131,20 +131,27 @@ abstract contract BaseStrategy {
...
+
+    // The maximum number of blocks between harvest calls
+    // NOTE: Override this value with your own, or set dynamically below
+    uint256 public maxReportDelay = 6300; // ~ once a day
 
-    // Adjust this to keep some of the position in reserve in the strategy,
+    // The minimum multiple that `callCost` must be above the credit/profit to be "justifiable"
+    // NOTE: Override this value with your own, or set dynamically below
+    uint256 public profitFactor = 100;
```
Macarse, line 142: shouldn't this be `virtual`?

contracts/BaseStrategy.sol
Macarse: there were basically two `expectedReturn()`, one in the vault, and one in the strategy.
There was no real use for the strategy's one.

contracts/BaseStrategy.sol
```diff
@@ -116,7 +116,7 @@ abstract contract BaseStrategy {
 
     // Version of this contract's StrategyAPI (must match Vault)
     function apiVersion() public pure returns (string memory) {
-        return "0.1.2";
+        return "0.1.3";
```
Macarse, line 119: can't you do `Vault.API_VERSION`?

contracts/BaseStrategy.sol
fubuloubu: Emergency shutdown also shunts the debt to `totalDebt`, so it would also be a step adjustment.

But regardless, I think having it be `virtual` is a worthwhile improvement. Can't possibly account for _every_ condition in the template haha.

contracts/BaseStrategy.sol
fubuloubu: Duh lol

contracts/BaseStrategy.sol
fubuloubu: Yeah, having the Vault maintain that statistic as "realized returns" seemed more appropriate.

contracts/BaseStrategy.sol
fubuloubu: We could add, but then we'd be forcing it to use one typical pricing function. Better to let the external user define that.

contracts/BaseStrategy.sol
fubuloubu: Very true, we can use the same trick with the gain and the return factor

contracts/BaseStrategy.sol
fubuloubu: That's a bigger change I hadn't thought about yet. Accepting suggestions though

contracts/BaseStrategy.sol
fubuloubu: No don't think it would work like that

contracts/BaseStrategy.sol
Grandthrax: I must be misunderstanding what you mean by step adjustment

fubuloubu: A step goes from `a` to `b ` instananeously (e.g. `y = b if t > 0 else a`)

A ramp goes from `a` to `b` incrementally (e.g. `y = m t + a`)

With the rate limit enabled, the total debt increases incrementally (ramps up) towards the debt limit. However, if the debt limit is lower than the total debt, it's called back all at once, letting the strategy decide the rate at which the debt is repaid.

Grandthrax: gotcha. I saw the debt rate but didn't realise it applied on the way out as well as the way in

fubuloubu: Not sure what is meant by "the way in", but the rate limit only applies when debt is being accumulated by strategies.

Where does that leave us w.r.t. this comment?

contracts/BaseStrategy.sol
```diff
@@ -241,6 +241,15 @@ abstract contract BaseStrategy {
      */
     function exitPosition() internal virtual;
 
+    /*
+     * Vault calls this function after shares are created during `Vault.report()`.
+     * You can customize this function to any share distribution mechanism you want.
+     */
+    function distributeRewards(uint256 _shares) external virtual {
```
dmolina79, line 248: if the strategist doesnt override this function all shares would be sent to the strategist?

fubuloubu: Yes, exactly! Current behavior today. Or, override and split to multiple parties. Unwrap some and split to another token. Burn it all! Possibilities are endless!

dmolina79 (approved):
LGTM!

contracts/BaseStrategy.sol
fubuloubu: fixed in cb2efcf

contracts/BaseStrategy.sol
fubuloubu: fixed in cb2efcf

contracts/test/TestStrategy.sol
fubuloubu: `BaseStrategy` has an example already. `TestStrategy` isn't really meant to be used as an example.

contracts/BaseStrategy.sol
fubuloubu: Fixed in 4c9d27c

contracts/BaseStrategy.sol
```diff
@@ -94,11 +94,11 @@ interface StrategyAPI {
 
     function keeper() external view returns (address);
 
-    function tendTrigger(uint256 gasCost) external view returns (bool);
+    function tendTrigger(uint256 callCost) external view returns (bool);
```
lbertenasco, line 97:
```suggestion
    function tendTrigger(uint256 _callCost) external view returns (bool);
```
Should this be underscored? (to be equal to other parameters)

PR #51, opened 2020-10-31
Add increase/decrease allowance functions

fubuloubu:
Since we removed the mitigation on approvals in #44, it makes sense to add the recommended mitigation which is 2 functions that atomically change the allowance

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -240,6 +240,36 @@ def approve(_spender : address, _value : uint256) -> bool:
     return True
 
 
+@external
+def increaseAllowance(_spender : address, _value : uint256) -> bool:
+    """
+    @dev Increase the allowance of the passed address to spend the total amount of tokens
+         on behalf of msg.sender. This method mitigates the risk that someone may use both
+         the old and the new allowance by unfortunate transaction ordering.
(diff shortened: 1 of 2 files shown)


fubuloubu:
Just 1 review and can merge

banteg (approved):
Looks good

chriswessels (approved):
LGTM!

contracts/Vault.vy
```diff
@@ -240,6 +240,36 @@ def approve(_spender : address, _value : uint256) -> bool:
     return True
 
 
+@external
+def increaseAllowance(_spender : address, _value : uint256) -> bool:
```
Macarse, line 244: why are you making the method signatures return bool if they always return True?

Macarse (approved):

contracts/Vault.vy
fubuloubu: It's what OZ uses https://github.com/OpenZeppelin/openzeppelin-contracts/blob/85334995073fff746961155113b0906b689a044c/contracts/token/ERC20/ERC20.sol#L170

Returning a value is actually cheaper for contracts because it doesn't have to perform a check that the call completed successfully, only the result of the call is True

PR #58, opened 2020-11-03
Add sweep amount

banteg:
This allows recovering less than full balance of ERC20 and also adds an ability to recover ERC721.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1274,7 +1274,7 @@ def erc20_safe_transfer(_token: address, _to: address, _value: uint256):
 
 
 @external
-def sweep(_token: address):
+def sweep(_token: address, _value: uint256 = MAX_UINT256):
     """
     @notice
         Removes tokens from this Vault that are not the type of token managed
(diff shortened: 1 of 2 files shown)


contracts/Vault.vy
```diff
@@ -1260,7 +1260,7 @@ def erc20_safe_transfer(_token: address, _to: address, _value: uint256):
 
 
 @external
-def sweep(_token: address):
+def sweep(_token: address, _value: uint256):
```
fubuloubu, line 1263: Would be neat to do the default value of `MAX_UINT256` (which just does balanceOf)

fubuloubu (commented):
Might be good to do an ERC721 test case... Somehow (or maybe not)

tests/functional/vault/test_misc.py
```diff
@@ -26,6 +26,7 @@ def test_sweep(gov, vault, rando, token, other_token):
         vault.sweep(other_token, {"from": rando})
 
     before = other_token.balanceOf(vault)
+    vault.sweep(other_token, 1, {"from": gov})
```
fubuloubu, line 29: Add an assertion after here lol

fubuloubu (approved):

tests/functional/vault/test_misc.py
fubuloubu:
```suggestion
    vault.sweep(other_token, 1, {"from": gov})
    assert other_token.balanceOf(vault) == balance - 1
    assert other_token.balanceOf(gov) == 1
```

tests/functional/vault/test_misc.py
```diff
@@ -26,6 +26,9 @@ def test_sweep(gov, vault, rando, token, other_token):
...
+    assert other_token.balanceOf(vault) == balance - 1
```
fubuloubu, line 30:
```suggestion
    assert other_token.balanceOf(vault) == before - 1
```

fubuloubu (approved):

Issue #45, opened 2020-10-27
Add Registry contract


lbertenasco:
> Work by submitting a PR to the [`feat/registry`](../tree/feat/registry) branch.

Contract is here: [`Registry.sol`](../blob/feat/registry/contracts/Registry.sol)

Let's continue [this discussion](https://github.com/iearn-finance/yearn-protocol/issues/58) on this issue.

freightledger:
Is there a pattern for the value of `vault api version` i.e. a regex or some reference to versioning? 

E.g. v1.1.0 != v1.01.0 where 1.1.0 is actually 1.10.1?

sambacha:
Is there a pattern for the value of `vault api version` i.e. a regex or some reference to versioning? 

E.g. v1.1.0 != v1.01.0 where 1.1.0 is actually 1.10.1?

PR #62, opened 2020-11-04
Vault Registry (take 2)

fubuloubu:
fixes: #45

Vault Registry, used for managing the list of "official" Yearn Vaults for the ecosystem.

Features:
- [x] Governance controlled (2-phase upgrade)
- [x] Getter for data of specific Vault
- [x] Getter for data of all supported Vaults (past and present)
- [x] Getter for latest Vault address for token

TODO:
- [ ] Tests
- [ ] Docs

### DIFF
--- a/contracts/Registry.sol
+++ b/contracts/Registry.sol
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.6.12;
+pragma experimental ABIEncoderV2;
+
+import "@openzeppelinV3/contracts/math/SafeMath.sol";
+import "@openzeppelinV3/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelinV3/contracts/utils/Address.sol";
+import "@openzeppelinV3/contracts/utils/EnumerableSet.sol";
+
+import {VaultAPI} from "./BaseStrategy.sol";
+
+struct VaultData {
+    address vaultAddress;
+    string vaultName;
+    string vaultSymbol;
+    string vaultApiVersion;
+    address tokenAddress;
+    string tokenName;
+    string tokenSymbol;
+    uint256 tokenDecimals; // NOTE: Vault.decimals() == Token.decimals()
+}
+
+contract Registry {
+    using Address for address;
+    using SafeMath for uint256;
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    address public governance;
+    address pendingGovernance;
+
+    EnumerableSet.AddressSet private vaults;
+
+    constructor(address _governance) public {
+        require(_governance != address(0), "Missing Governance");
+        governance = _governance;
+    }
+
+    // Governance setters
+    function setGovernance(address _governance) external onlyGovernance {
+        pendingGovernance = _governance;
+    }
+
+    function acceptGovernance() external {
+        require(msg.sender == pendingGovernance, "!pendingGovernance");
+        governance = msg.sender;
+    }
+
+    modifier onlyGovernance {
+        require(msg.sender == governance, "!governance");
+        _;
+    }
+
+    // Vault set actions
+    function addVault(address _vault) public onlyGovernance {
+        require(_vault.isContract(), "Vault is not a contract");
+        // Checks if vault is already on the array
+        require(!vaults.contains(_vault), "Vault already exists");
+        // Adds unique _vault to vaults array
+        vaults.add(_vault);
+    }
+
+    function removeVault(address _vault) public onlyGovernance {
+        // Checks if vault is already on the array
+        require(vaults.contains(_vault), "Vault not in set");
+        // Remove _vault to vaults array
+        vaults.remove(_vault);
+    }
+
+    function getVaultData(address _vault) internal view returns (VaultData memory) {
+        address token = VaultAPI(_vault).token();
+        return
+            VaultData({
+                vaultAddress: _vault,
+                vaultName: VaultAPI(_vault).name(),
+                vaultSymbol: VaultAPI(_vault).symbol(),
+                vaultApiVersion: VaultAPI(_vault).apiVersion(),
+                tokenAddress: token,
+                tokenName: ERC20(token).name(),
+                tokenSymbol: ERC20(token).symbol(),
+                tokenDecimals: ERC20(token).decimals()
+            });
+    }
+
+    // Vaults getters
+    function getVault(uint256 _index) external view returns (VaultData memory) {
+        return getVaultData(vaults.at(_index));
+    }
+
+    function getVault(address _vault) external view returns (VaultData memory) {
+        return getVaultData(_vault);
+    }
+
+    function getVaultsLength() external view returns (uint256) {
+        return vaults.length();
+    }
+
+    function getVaults() external view returns (VaultData[] memory) {
+        VaultData[] memory vaultsArray = new VaultData[](vaults.length());
+        for (uint256 i = 0; i < vaults.length(); i++) {
+            vaultsArray[i] = getVaultData(vaults.at(i));
+        }
+        return vaultsArray;
+    }
+}
--- a/contracts/Registry.sol
+++ b/contracts/Registry.sol
@@ -29,6 +29,9 @@ contract Registry {
     address pendingGovernance;
 
     EnumerableSet.AddressSet private vaults;
+    mapping(address => address) public latestVaultForToken;
+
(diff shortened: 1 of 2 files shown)


lbertenasco:
would it be possible to have an array of active strategies added to ` VaultData`?

milkyklim (approved):

fubuloubu:
> would it be possible to have an array of active strategies added to ` VaultData`?

That's a lot of extra data to manage. There might be 5-20 strategies per Vault, so I don't think there's any reasonable way to represent this in a UI. I never understood why we added it originally, probably because the 1 strategy per Vault limitation made it the easiest way to see what's happening with the Vault.

I would be much more in favor of adding aggregate statistics instead, such as:
- `Vault.totalAssets()` (current assets in the Vault and connected Strategies)
- `Vault.maxAvailableShares()` (current withdrawal size)
- `Vault.pricePerShare()` (cumulative returns)
- `Vault.debtLimit()` (sum of all connected strategy debt limits)
- `Vault.depositLimit()` (if there is a limit on the overall deposits this Vault will take)
- `Vault.totalDebt()` (current debt lent out to all connected Strategies)
- `Vault.managementFee()` (Governance Fee for management of Vault)
- `Vault.performanceFee()` (Governance Fee for performance of Vault)

Also, maybe we can explore adding 12d/50d EMA for at least `Vault.pricePerShare()` (if not some of the others) to get an on-chain trend for Vault APY, which could be very useful for other integrating projects. This would require a continuous update mechanism across all the Vaults, probably a good candidate for Keep3r integration

lbertenasco:
RegistryV1 only purpose was to allow getting the active strategy for a vault.
It was a bit complex for devs to find the current strategy of a vault specially because we had regular, delegated and wrapped types.

I know x48 and others need to know the strategies of each vault for the API + graph + webapps.

Yes! It'd be awesome if we could add a few statistics :)

My concern is that we don't have a way to get all vault strategies with a single call.
Right now we rely on these to know active strategies for V2:
```sol
event StrategyAdded:
strategies: public(HashMap[address, StrategyParams])
withdrawalQueue: public(address[MAXIMUM_STRATEGIES]) # fastest query but not reliable
```

I can add a strategies array on `Vault.py` that keeps track of "connected/active" strategies only.
This would help a LOT :)

WDYT?

fubuloubu:
> RegistryV1 only purpose was to allow getting the active strategy for a vault.

I just really push back on why this is needed. What is the point/idea behind getting the currently active strategy with v1? What does it give you?

My theory here is that it gives you a proxy for the same kinds of statistics we're actually looking for the Vault to produce. With the v2 design, it's now behind a layer of abstraction, for good reason! It will be difficult to keep track of all of these connected strategies without being involved in their development and maintenance. I think most end users and integrators are interested in higher level statistics about the Vault, rather than the low level details of how each connected strategy is doing.

---

> I can add a strategies array on `Vault.py` that keeps track of "connected/active" strategies only.

Actually, the right place for this would be the Registry here. Keep a mapping of each Vault's set of the Strategies with a non-zero balance sheet, and then call `Vault.totalBalanceSheet(vaultStrategies[_vault])` to get a Vault's entire "live" balance sheet for real-time calls. (standard disclaimer about using estimated values from off-chain systems applies)

listonjesse:
> I just really push back on why this is needed. What is the point/idea behind getting the currently active strategy with v1? What does it give you?
> 
> My theory here is that it gives you a proxy for the same kinds of statistics we're actually looking for the Vault to produce. With the v2 design, it's now behind a layer of abstraction, for good reason! It will be difficult to keep track of all of these connected strategies without being involved in their development and maintenance. I think most end users and integrators are interested in higher level statistics about the Vault, rather than the low level details of how each connected strategy is doing.

Right now the only method I can think of to figure a vault's active strategies is by consuming all events from the contract creation and until the current block and them building up that list as you find strat add and strat remove events. This seems doable for any dev integrations that truly want to know the current state of strategies. 

Is there an on-chain use case that would benefit from this info being available from a contract? Will vaults ever need maintainence done via keep3rs (for example, could we ever imagine a keep3r job that loops through the active strategies to call some tx on all of them). 

In v1, knowing the strategies was super useful to the informed user. 

You could figure out if withdrawals were free by figuring how much of a vault wasn't yet deployed to a strategy.

You could very easily dig into each strat's implementation details and figure out how much pending want token was ready to be harvested if you were trying to time deposits into vaults with more pending harvest than othere.

You could also use the strat to figure out if a vault was 'effectively' disabled by seeing if the strat was completely empty or heavily 'diluted' by funds that cannot be moved to the strategy, see TUSD and DAI vaults today for an example of this.

If v2 truly abstracts away the need for the user to know about strategies and eliminates any sort of advantages that comes with knowing these details, then that exites me.

fubuloubu:
> If v2 truly abstracts away the need for the user to know about strategies and eliminates any sort of advantages that comes with knowing these details, then that exites me.

@listonjesse I wouldn't necessarily say that it eliminates any sort of advantages that comes with knowing intimate details of the Vault's connected Strategies (because there is always some advantage to having more detailed information), but I would say that the added complexity to support people trying to time deposits and withdrawals isn't worth it. The intended use case of the Vault is "set it and forget it" kind of behavior, where longer term trends should be of much more importance than trying to game the system with shorter term deposits.

---

> You could figure out if withdrawals were free by figuring how much of a vault wasn't yet deployed to a strategy.

We have `Vault.maxAvailableShares()` for this (withdrawable depth).

> You could ... figure out how much pending want token was ready to be harvested

Harvests should be more regularly occurring and better balanced, the 12d EMA of `Vault.pricePerShare()` should give better information here (including if you charted it to notice any recurring trends you'd want to play).

> You could also use the strat to figure out if a vault was 'effectively' disabled

The ratio of `Vault.totalDebt()` to `Vault.debtLimit()` should tell you something of this nature. Also, `Vault.emergencyShutdown()` if it's *actually* shut off.

lbertenasco (approved):
wild vault types appear :)

contracts/Registry.sol
```diff
@@ -0,0 +1,182 @@
...
+        address token = VaultAPI(_vault).token();
+        address oldVault = latestVaultForToken[token];
+        latestVaultForToken[token] = _vault;
+        emit VaultUpgrade(token, oldVault, _vault);
+    }
+
+    function endorseVault(address _vault) public onlyGovernance {
+        require(experimentalVaults.contains(_vault), "!experimental");
+        experimentalVaults.remove(_vault);
+        endorsedVaults.add(_vault);
```
salazarguille, line 90: I think you should emit an event after adding the vault in the endorsedVaults list. Especially when you move the vault from "experimental" to "endorsed".

contracts/Registry.sol
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    function removeVault(address _vault) public onlyGovernance {
+        // Checks if vault is already on the array
+        require(vaults.contains(_vault), "Vault not in set");
+        // Remove _vault to vaults array
+        vaults.remove(_vault);
```
salazarguille, line 97: Do you need to have the vaults in two different lists? Maybe the status can be done using a mapping, so then you only need one list for the vault. 

```mapping(address => VaultStatus) public  status;```

contracts/Registry.sol
salazarguille, line 90: I think you should emit an event after adding the vault in the endorsedVaults list. Especially when you move the vault from "experimental" to "endorsed".

salazarguille (changes requested):

contracts/Registry.sol
fubuloubu: I wanted to add a status enum to each Vault, but the end goal here is to make it easy for UI integration. From a UI perspective, it wants to render the list of Vaults, depending on what the user cases about.

A user cares about a few things:
1. Do I need to "upgrade" any of my Vault shares (`getDeprecatedVaults()` migrate to `getEndorsedVaults()`)
2. Do I want to use any of the Vaults in "beta" (`getExperimentalVaults()`)
3. What are my current holdings (`getActiveVaults()`)

Being able to view them in a list format works better for JS, otherwise we'd be maintaining an Enum in two separate places. We'd also have to perform an interation over a series of calls that's avoided from having these lists. It's definitely not the most efficient thing in the world, but there's some method to the madness.

contracts/Registry.sol
fubuloubu: Agree, status changes should be tracked via events

fubuloubu:
Closing in favor of more simplified approach e.g. https://github.com/iearn-finance/yearn-vaults/pull/124

PR #64, opened 2020-11-06
Fix weird intermittent test failure in CI

fubuloubu:
Because it's annoying and I hate it

### DIFF
--- a/tests/functional/strategy/test_misc.py
+++ b/tests/functional/strategy/test_misc.py
@@ -2,7 +2,7 @@ import pytest
 import brownie
 
 
-def test_harvest_tend_authority(gov, keeper, strategist, strategy, rando, fn_isolation):
+def test_harvest_tend_authority(gov, keeper, strategist, strategy, rando):
     # Only keeper, strategist, or gov can call tend
     strategy.tend({"from": keeper})
     strategy.tend({"from": strategist})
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -82,18 +80,16 @@ jobs:
...
-          ~/.vvm
-        key: ${{ runner.os }}-compiler-cache
+          ~/.brownie
+          ./build
+        key: build-cache
 
     - name: Setup node.js
       uses: actions/setup-node@v1
-      with:
-        node-version: '12.x'
```
Macarse, line 96: why did you remove this? I don't know what's the default, but let's have a hardcoded version at least so it's reproducible.

.github/workflows/test.yaml
```diff
@@ -82,18 +80,16 @@ jobs:
...
-    - name: Cache compiler installations
+    - name: Cache Brownie Build Folders
       uses: actions/cache@v2
       with:
         path: |
-          ~/.solcx
-          ~/.vvm
-        key: ${{ runner.os }}-compiler-cache
+          ~/.brownie
+          ./build
```
Macarse, line 88: with this, ci would compile only the diff.

.github/workflows/test.yaml
```diff
@@ -82,18 +80,16 @@ jobs:
           ~/.vvm
         key: compiler-cache
 
-    - name: Cache compiler installations
+    - name: Cache Brownie Build Folders
       uses: actions/cache@v2
       with:
         path: |
-          ~/.solcx
-          ~/.vvm
```
Macarse, line 90: why did you remove these? they make sense

.github/workflows/test.yaml
fubuloubu: That's not even a hardcoded version, it just uses 14.x by default

.github/workflows/test.yaml
fubuloubu: Yep!

.github/workflows/test.yaml
fubuloubu: it's in 2 places

Macarse (approved):

PR #66, opened 2020-11-06
Forgot to assert bool returns on token transfers

fubuloubu:


spinoch:
Could you add a test for this?

contracts/Vault.vy
```diff
@@ -791,7 +791,7 @@ def withdraw(_shares: uint256 = MAX_UINT256, _recipient: address = msg.sender) -
     log Transfer(msg.sender, ZERO_ADDRESS, shares)
```
spinoch, line 791: Wouldn't you want to assert this also?

fubuloubu: logs appear in the transaction receipt. broadly, we should probably test for event logs occurring correctly, but haven't gotten to that yet.

fubuloubu:
> Could you add a test for this?

There's a series of weird token patterns that this code will revert on, so it's important that the deployer understands the token that they are using for their vault

eosioAdminmanager (approved):
@0xa60e9d48477bf0fd37fa2857aa904c09e6d26444

eosioAdminmanager (approved):

contracts/Vault.vy
spinoch: How about changing this to
```suggestion
    assert self.transfer(ZERO_ADDRESS, shares)
```
?

spinoch: And if I remember correctly there's another place where `log Transfer` is used and this could also be done

fubuloubu: Ah, I think I see the confusion. This function is taking place of a mint function, it doesn't actually transfer any shares but creates them. This is a weird aspect of using ERC20.

spinoch: Ah, right. That somewhat makes sense

lbertenasco (approved):

PR #68, opened 2020-11-06
Separate strategist and rewards roles

banteg:


contracts/BaseStrategy.sol
```diff
@@ -176,6 +178,11 @@ abstract contract BaseStrategy {
         keeper = _keeper;
     }
 
+    function setRewards(address _rewards) external {
+        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
```
fubuloubu, line 182: Seems like something governance shouldn't need rights to? They can always change `strategist` to themselves, then change this
```suggestion
        require(msg.sender == strategist, "!authorized");
```

fubuloubu (approved):

tests/functional/strategy/test_config.py
```diff
@@ -45,6 +46,7 @@ def test_strategy_setEmergencyExit(strategy, gov, strategist, rando, chain):
     "getter,setter,val",
     [
         ("strategist", "setStrategist", None),
+        ("rewards", "setRewards", None),
```
fubuloubu, line 49: Shit, the way this test works, you should skip governance trying to set it

banteg: Fixed by adding `caller` into parametrize setup.

tests/functional/strategy/test_config.py
```diff
@@ -64,13 +66,10 @@ def test_strategy_setParams(gov, strategist, strategy, rando, getter, setter, va
...
         getattr(strategy, setter)(val, {"from": rando})
 
-    getattr(strategy, setter)(val, {"from": strategist})
-    assert getattr(strategy, getter)() == val
+    caller = {"gov": gov, "strategist": strategist}[caller]
 
-    if getter == "strategist":
-        # Strategist can't change themselves once they update
-        with brownie.reverts():
-            strategy.setStrategist(strategist, {"from": strategist})
```
fubuloubu, line 71: Still import to check this condition, but could do in another test case I suppose

fubuloubu (approved):

PR #69, opened 2020-11-07
Add Strategy and Vault Yield Oracles

fubuloubu:
This feature adds two new parameters to all connected Strategies: 12-day and 50-day EMAs of the Strategy's annual returns

It also adds a debt-weighted sum of the Vault's overall annual return, using the 12-day and 50-day EMAs of all connected Strategies

It isn't known how well these will perform in practice. However, tests were added that ensures it should not generate any unexpected issues. Further experimentation is required.

TODO:
- [ ] Add functional tests
- [x] Add integration tests


fubuloubu:
Will have to re-approach this later

Issue #49, opened 2020-10-29
Add a way to report realized losses


banteg:
Example with `StrategyUniswapPairPickle` migration. Strategy has incurred a 0.5% loss caused by Pickle withdrawal fee, but the `pricePerShare` has remained the same, meaning the vault is actually short that amount.
```
pricePerShare 1.000001852497160145
estimatedTotalAssets 15.728413834807870538
Transaction sent: 0x8f5150e727b70961708548cad44ef527f48d446ec1f18813f2f2c6e786b22cd9
  Gas price: 0.0 gwei   Gas limit: 10000000
  Vault.migrateStrategy confirmed - Block: 11152421   Gas used: 251525 (2.52%)

pricePerShare 1.000001852497160145
estimatedTotalAssets 15.648645586618978664
```

PR #74, opened 2020-11-13
Add ability to report losses

fubuloubu:
fixes: #49

TODO:
- [x] tests
- [ ] docs

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1228,17 +1228,18 @@ def report(_gain: uint256, _loss: uint256) -> uint256:
     # NOTE: This must be done at the same time, to ensure the relative
     #       ratio of governance_fee : strategist_fee is kept intact
     total_fee: uint256 = governance_fee + strategist_fee
-    reward: uint256 = self._issueSharesForAmount(self, total_fee)
+    if total_fee > 0:
+        reward: uint256 = self._issueSharesForAmount(self, total_fee)
 
-    # Send the rewards out as new shares in this Vault
-    if strategist_fee > 0:  # NOTE: Guard against DIV/0 fault
-        # NOTE: Unlikely to throw unless sqrt(reward) >>> 1e39
-        strategist_reward: uint256 = (strategist_fee * reward) / total_fee
-        self._transfer(self, msg.sender, strategist_reward)
-        Strategy(msg.sender).distributeRewards(strategist_reward)
-    # NOTE: Governance earns any dust leftover from flooring math above
-    if self.balanceOf[self] > 0:
-        self._transfer(self, self.rewards, self.balanceOf[self])
+        # Send the rewards out as new shares in this Vault
+        if strategist_fee > 0:  # NOTE: Guard against DIV/0 fault
+            # NOTE: Unlikely to throw unless sqrt(reward) >>> 1e39
+            strategist_reward: uint256 = (strategist_fee * reward) / total_fee
+            self._transfer(self, msg.sender, strategist_reward)
+            Strategy(msg.sender).distributeRewards(strategist_reward)
+        # NOTE: Governance earns any dust leftover from flooring math above
+        if self.balanceOf[self] > 0:
+            self._transfer(self, self.rewards, self.balanceOf[self])
 
     # Compute the line of credit the Vault is able to offer the Strategy (if any)
     credit: uint256 = self._creditAvailable(msg.sender)
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -90,7 +90,8 @@ struct StrategyParams:
     rateLimit: uint256  # Increase/decrease per block
     lastReport: uint256  # block.number of the last time a report occured
     totalDebt: uint256  # Total outstanding debt that Strategy has
-    totalReturns: uint256  # Total returns that Strategy has realized for Vault
+    totalGain: uint256  # Total returns that Strategy has realized for Vault
+    totalLoss: uint256  # Total losses that Strategy has realized for Vault
 
 
 event StrategyAdded:
@@ -102,10 +103,12 @@ event StrategyAdded:
 
 event StrategyReported:
     strategy: indexed(address)
-    returnAdded: uint256
-    debtAdded: uint256
-    totalReturn: uint256
+    gain: uint256
+    loss: uint256
+    totalGain: uint256
+    totalLoss: uint256
     totalDebt: uint256
+    debtAdded: uint256
     debtLimit: uint256
 
 
@@ -858,7 +861,8 @@ def addStrategy(
         rateLimit: _rateLimit,
         lastReport: block.number,
         totalDebt: 0,
-        totalReturns: 0,
+        totalGain: 0,
+        totalLoss: 0,
     })
     self.debtLimit += _debtLimit
     log StrategyAdded(_strategy, _debtLimit, _rateLimit, _performanceFee)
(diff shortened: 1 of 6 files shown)


contracts/Vault.vy
```diff
@@ -1166,20 +1168,35 @@ def report(_return: uint256) -> uint256:
...
+        last report, and should be accounted for on the Vault's balance sheet
     @return Amount of debt outstanding (iff totalDebt > debtLimit).
     """
 
     # Only approved strategies can call this function
     assert self.strategies[msg.sender].activation > 0
 
+    # We have a loss to report, do it before the rest of the calculations
+    if _loss > 0:
+        loss: uint256 = min(_loss, self.strategies[msg.sender].totalDebt)
```
dmolina79, line 1190: gain and loss are supposed to be mutually exclusive? meaning if i have a loss to report there shouldn't be any gain on that call, unless im missing something.  

Just thinking if we need to require if loss > 0 gain should equal 0 and vice versa, only issue is that the strategy could mess up and get stuck failing to not report

contracts/Vault.vy
```diff
@@ -1228,10 +1245,10 @@ def report(_return: uint256) -> uint256:
     # the credit increase we are offering (if any)
     # NOTE: This is just used to adjust the balance of tokens between the Strategy and
     #       the Vault based on the Strategy's debt limit (as well as the Vault's).
-    if _return < credit:  # credit surplus, give to Strategy
-        assert self.token.transfer(msg.sender, credit - _return)
-    elif _return > credit:  # credit deficit, take from Strategy
-        assert self.token.transferFrom(msg.sender, self, _return - credit)
+    if _gain < credit:  # credit surplus, give to Strategy
+        assert self.token.transfer(msg.sender, credit - _gain)
```
dmolina79, line 1249: `_gain` could be zero, so we would give back completed credit amount right?

contracts/Vault.vy
fubuloubu: I left it open here. I could see a scenario where you had both a gain and a loss to report e.g. you cashed in some reward tokens, but you lost some amount of underlying in the process. Of course, you could just mix the two together, but I think leaving it open makes it more flexible.

contracts/Vault.vy
fubuloubu: Yes, that's the idea

contracts/Vault.vy
```diff
@@ -1273,7 +1290,7 @@ def report(_return: uint256) -> uint256:
 
     log StrategyReported(
         msg.sender,
-        _return,
+        _gain,
```
dmolina79, line 1293: do we want to track reported loss value in event as well?

dmolina79 (approved):
LGTM!, recommend a second review from @banteg since he requested the change

contracts/Vault.vy
fubuloubu: Yes, I'll probably change up the events in a later PR too

contracts/Vault.vy
```diff
@@ -1155,7 +1157,7 @@ def expectedReturn(_strategy: address = msg.sender) -> uint256:
 
 
 @external
-def report(_return: uint256) -> uint256:
+def report(_gain: uint256, _loss: uint256) -> uint256:
```
dmolina79, line 1160: is it a wash if gain and loss are reported by any chance with same number?

fubuloubu: A reported loss gets accounted for first, reducing debt, so it'll look like you earned more with less I suppose.

contracts/Vault.vy
```diff
@@ -102,10 +103,12 @@ event StrategyAdded:
...
-    returnAdded: uint256
-    debtAdded: uint256
-    totalReturn: uint256
+    gain: uint256
+    loss: uint256
+    totalGain: uint256
+    totalLoss: uint256
     totalDebt: uint256
+    debtAdded: uint256
     debtLimit: uint256
```
fubuloubu, line 112: @banteg made a big change here

tests/functional/vault/test_losses.py
```diff
@@ -0,0 +1,46 @@
...
+    token.approve(vault, 500, {"from": gov})
+    vault.deposit(500, {"from": gov})
+
+    strategy.harvest({"from": gov})
+    assert token.balanceOf(strategy) == 500
+
+    # First loss
+    strategy._takeFunds(100, {"from": gov})
+    strategy.harvest({"from": gov})
+    params = vault.strategies(strategy).dict()
```
fubuloubu, line 28: We have to do this to all the tests :smiling_imp:

fubuloubu: https://github.com/iearn-finance/yearn-vaults/pull/76

tests/functional/vault/test_strategies.py
```diff
@@ -20,7 +20,7 @@ def test_addStrategy(web3, gov, vault, strategy, rando):
     with brownie.reverts():
         vault.addStrategy(strategy, 1000, 10, 50, {"from": rando})
 
-    assert vault.strategies(strategy) == [0, 0, 0, 0, 0, 0, 0]
+    assert vault.strategies(strategy) == [0, 0, 0, 0, 0, 0, 0, 0]
```
banteg, line 23: With so many params, should we switch to comparing dicts?

banteg (approved):
Looks great, this opens up a whole new world of possibilities!

tests/functional/vault/test_strategies.py
fubuloubu: See #76 (which I didn't know existed until today, lol)

contracts/Vault.vy
dmolina79: makes sense, would it be good to add a test case to check values after side effects?

fubuloubu: We added a test case in 1cb83d1, do you think it can be improved?

dmolina79: yeah, maybe we can add a mock strategy that sends bad signals in testing like gain=100K and loss=100K just to check the state of the vault @Macarse, would that make sense?

dmolina79: Ill add it as an issue to track to improve testing, just for completion sake i would say, glad to take a stab at it

contracts/Vault.vy
```diff
@@ -1155,7 +1159,7 @@ def expectedReturn(_strategy: address = msg.sender) -> uint256:
 
 
 @external
-def report(_return: uint256) -> uint256:
+def report(_gain: uint256, _loss: uint256) -> uint256:
```
Grandthrax, line 1162: Gain is the wrong word to use. Gain implies that the number is profit. But that is not the case. The number is profit + debtOutstanding

contracts/Vault.vy
```diff
@@ -1166,20 +1170,35 @@ def report(_return: uint256) -> uint256:
...
         reserves. Anything outside of those bounds is abnormal behavior.
 
         All approved strategies must have increased diligence around
         calling this function, as abnormal behavior could become catastrophic.
-    @param _return
-        Amount Strategy has made on it's investment since its last report,
-        and is free to be given back to Vault as earnings
+    @param _gain
+        Amount Strategy has realized as a gain on it's investment since its
+        last report, and is free to be given back to Vault as earnings
```
Grandthrax, line 1180: Plus any debt outstanding

contracts/Vault.vy
```diff
@@ -1166,20 +1170,35 @@ def report(_return: uint256) -> uint256:
...
     """
 
     # Only approved strategies can call this function
     assert self.strategies[msg.sender].activation > 0
 
+    # We have a loss to report, do it before the rest of the calculations
+    if _loss > 0:
+        loss: uint256 = min(_loss, self.strategies[msg.sender].totalDebt)
+        self.strategies[msg.sender].totalLoss += loss
+        self.strategies[msg.sender].totalDebt -= loss
```
Grandthrax, line 1194: You either need to decrease from totalGain as well or change the _expectedReturn method

contracts/Vault.vy
```diff
@@ -1124,14 +1128,14 @@ def creditAvailable(_strategy: address = msg.sender) -> uint256:
 def _expectedReturn(_strategy: address) -> uint256:
```
Grandthrax, line 1128: if totalGain does not include losses then you need to update this method. See my comment further down where you calculate loss in report() method

contracts/Vault.vy
Grandthrax, line 1162: Please can we have a second report method with 1 input that calls this with loss = 0. So that this vault upgrade will still work with legacy strategies

Grandthrax (commented):
mostly looks good. Some clarification should be added to show the user how to deal with debt outstanding.

contracts/Vault.vy
fubuloubu: You can't migrate a strategy to a newer Vault, the solution here will be to upgrade and re-release the strategy.

fubuloubu: perhaps this should be revisited, have it be simply reported gains, but then add a `debtPayment` field too

Grandthrax: That would certainly make it easier for new users to understand.

PR #75, opened 2020-11-13
feat: test losses

Macarse:


fubuloubu:
Cherry picked into #74

PR #76, opened 2020-11-13
Refactor all tests to use dict returns for structs

fubuloubu:
Based on: #74

Why didn't I know this feature existed?

### DIFF
--- a/tests/functional/strategy/test_migration.py
+++ b/tests/functional/strategy/test_migration.py
@@ -8,9 +8,7 @@ def test_good_migration(
     strategy.harvest({"from": strategist})
 
     strategy_debt = vault.strategies(strategy).dict()["totalDebt"]
-    prior_position = token.balanceOf(strategy)
-    assert strategy_debt > 0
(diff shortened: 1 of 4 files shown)


tests/functional/strategy/test_migration.py
```diff
@@ -25,8 +25,8 @@ def test_good_migration(
         vault.migrateStrategy(strategy, new_strategy, {"from": guardian})
 
     vault.migrateStrategy(strategy, new_strategy, {"from": gov})
-    assert vault.strategies(strategy)[4] == 0
-    assert vault.strategies(new_strategy)[4] == strategy_debt
+    assert vault.strategies(strategy).dict()["totalDebt"] == 0
+    assert vault.strategies(new_strategy).dict()["totalDebt"] == strategy_debt
```
fubuloubu, line 29: Note that these were previously incorrect, so there is a latent failure in this test case

dmolina79 (approved):
LGTM! , big improvement in readability of the code

Issue #73, opened 2020-11-12
Add "Available Deposit Limit" as view-only function


fubuloubu:
Add a view function that provides the following:
```python
return self.depositLimit - self._totalAssets() if self.depositLimit >= self._totalAssets() else 0
```

PR #79, opened 2020-11-13
feat: add Available Deposit Limit as view-only function

Macarse:
fixes: #73

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1144,6 +1144,15 @@ def _expectedReturn(_strategy: address) -> uint256:
 
 @view
 @external
+def availableDepositLimit() -> uint256:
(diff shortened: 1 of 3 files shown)


contracts/Vault.vy
```diff
@@ -1142,6 +1142,15 @@ def _expectedReturn(_strategy: address) -> uint256:
         return 0  # Covers the scenario when block.number == strategy_activation
 
 
+@view
+@external
+def availableDepositLimit() -> uint256:
+    if self.depositLimit >= self._totalAssets():
```
fubuloubu, line 1148: I know I said otherwise in the issue, but if they're equal, we can save some gas by returning 0
```suggestion
    if self.depositLimit > self._totalAssets():
```

fubuloubu (approved):

Issue #71, opened 2020-11-10
Add Guest List


fubuloubu:
We might want to let Vaults to be able to define "guest lists", which would be another contract that dictates who is allowed to participate in a Vault (and transfer shares). Potential use case here might be for more "experimental" style Vaults, or for "high risk" Vaults where we might want to define a limit based on something like $ value of YFI staked in governance, etc.

This would look like the following:
```python
interface GuestList:
    def authorized(guest: address) -> bool: view

guestList: public(GuestList)


... # In deposit()
    if self.guestList.address != ZERO_ADDRESS:
         assert self.guestList.authorized(msg.sender)
...
```

WrongNebula:
Can we have Bounc3rs at the gate too?

ShadeUndertree:
I'd like to pick this one up.

fubuloubu:
Note: add "deposit size" to authorization check, so we can do limits based on total wealth, or amount of YFI, or stuff like that, e.g.:
```python
interface GuestList:
    def authorized(guest: address, amount: uint256) -> bool: view

guestList: public(GuestList)


... # In deposit(_amount)
    if self.guestList.address != ZERO_ADDRESS:
         assert self.guestList.authorized(msg.sender, _amount)
...
```

banteg:
I've implemented this concept here. A bouncer can maintain a guest list, as well as a list of tokens which guarantee entrance if a user holds a certain min amount of them.
https://github.com/banteg/guest-list/blob/master/contracts/GuestList.vy

ShadeUndertree:
@banteg would you like me to take what you've written and add it to the v2 vaults, plus tests? Or should I move on to something else? (since your implementation seems pretty comprehensive, I'd just be copying & pasting and adding a few tests)

fubuloubu:
For this feature, I'd say we add a "test guest list" that is just a simpler version of what @banteg implemented. His version can probably continue adding new features (maybe merkle tree claim check?)

PR #84, opened 2020-11-17
feat: add guest list (#71)

ShadeUndertree:
fixes: #71 

Co-authored-by: banteg <banteeg@gmail.com>
Co-authored-by: Just somie guy <3859395+fubuloubu@users.noreply.github.com>

### DIFF
--- a/tests/functional/vault/test_shares.py
+++ b/tests/functional/vault/test_shares.py
@@ -9,7 +9,7 @@ def vault(gov, token, Vault):
 
 
 @pytest.fixture
-def guestList(gov, TestGuestList):
+def guest_list(gov, TestGuestList):
     yield gov.deploy(TestGuestList)
 
 
@@ -27,18 +27,16 @@ def test_deposit_with_wrong_amount(vault, token, gov):
         vault.deposit(balance, {"from": gov})
 
 
-def test_deposit_with_guest_list_not_invited(
-    vault, guestList, token, gov, rando, history
-):
+def test_deposit_with_guest_list(vault, guest_list, token, gov, rando, history):
     # Make sure we're attempting to deposit something
     token.transfer(rando, token.balanceOf(gov) // 2, {"from": gov})
     balance = token.balanceOf(rando)
     token.approve(vault, balance, {"from": rando})
 
-    # Note - don't need to call guestList.setGuests, since nobody's invited by
+    # Note - don't need to call guest_list.setGuests, since nobody's invited by
     # default.
     # gov is our bouncer
-    vault.setGuestList(guestList, {"from": gov})
+    vault.setGuestList(guest_list, {"from": gov})
 
     # Ensure rando's not permitted to deposit
     with brownie.reverts():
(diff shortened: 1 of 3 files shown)


tests/functional/vault/test_shares.py
```diff
@@ -8,6 +8,11 @@ def vault(gov, token, Vault):
     yield gov.deploy(Vault, token, gov, gov, "", "")
 
 
+@pytest.fixture
+def guestList(gov, TestGuestList):
+    yield gov.deploy(TestGuestList)
```
fubuloubu, line 13: In python code use snake_case

tests/functional/vault/test_shares.py
```diff
@@ -22,6 +27,59 @@ def test_deposit_with_wrong_amount(vault, token, gov):
...
+    # Deposit balance
+    vault.deposit(balance, {"from": rando})
+
+    # Ensure authorized was called
+    authorizedCall = next(
+        sc
+        for sc in history[-1].subcalls
+        if sc["function"] == "authorized(address,uint256)"
+    )
+    assert authorizedCall["return_value"][0] == True
```
fubuloubu, line 76: This isn't really necessary, given the result of the deposit below

tests/functional/vault/test_shares.py
```diff
@@ -22,6 +27,59 @@ def test_deposit_with_wrong_amount(vault, token, gov):
...
+    with brownie.reverts():
+        vault.deposit(balance, {"from": rando})
+
+    # Ensure authorized was called and that the deposit didn't revert for a
+    # different reason.
+    assert history[-1].subcalls[-1]["function"] == "authorized(address,uint256)"
+    assert history[-1].subcalls[-1]["return_value"][0] == False
+
+
+def test_deposit_with_guest_list_invited(vault, guestList, token, gov, rando, history):
```
fubuloubu, line 53: Can merge this test with the other test by first showing it reverts when they're not authorized, then they can deposit when they are. The setup is basically the same.

fubuloubu (changes requested):
Minor feedback, otherwise looks good! Thank you!

tests/functional/vault/test_shares.py
ShadeUndertree: Makes perfect sense, will do.

tests/functional/vault/test_shares.py
ShadeUndertree: The trouble is that if the guestList.authorized check in the Vault (lines 671 & 672) are removed, this call still succeeds, presenting a false positive.

tests/functional/vault/test_shares.py
ShadeUndertree: Thanks, will fix!

tests/functional/vault/test_shares.py
ShadeUndertree: Nevermind, if I combine the test cases as you've suggested, that problem goes away since I check authorized was the cause of the revert.

PR #92, opened 2020-11-18
Upgrade Keeper Script

fubuloubu:
~~depends: https://github.com/eth-brownie/brownie/pull/847~~

### DIFF
--- a/scripts/keep.py
+++ b/scripts/keep.py
@@ -35,7 +35,9 @@ def main():
         assert (
             strategy.keeper() == bot.address
         ), "Bot is not set as keeper! [{strategy.address}]"
-        assert strategy.vault() == vault.address, "Vault mismatch! [{strategy.address}]"
+        assert (
+            strategy.vault() == vault.address
+        ), f"Vault mismatch! [{strategy.address}]"
 
     while True:
         starting_balance = bot.balance()
--- a/scripts/keep.py
+++ b/scripts/keep.py
@@ -1,8 +1,14 @@
-from brownie import accounts, network, interface, Vault
+from brownie import accounts, network, interface, Vault, Token
+from brownie.network.gas.strategies import GasNowScalingStrategy
 from decimal import Decimal
 from eth_utils import is_checksum_address
 import requests
 from time import sleep
+
+
+GAS_BUFFER = 1.2
+
+gas_strategy = GasNowScalingStrategy()
 
 
 def get_address(msg: str) -> str:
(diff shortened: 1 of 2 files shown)


fubuloubu:
Note to reviewers: must manually install the brownie branch listed above to run

scripts/keep.py
```diff
@@ -38,37 +38,69 @@ def main():
         assert strategy.vault() == vault.address, "Vault mismatch! [{strategy.address}]"
```
banteg, line 38:
```suggestion
        assert strategy.vault() == vault.address, f"Vault mismatch! [{strategy.address}]"
```

scripts/keep.py
```diff
@@ -38,37 +38,69 @@ def main():
...
             debt = vault.debtOutstanding(strategy) / 10 ** vault.decimals()
             print(f"[{strategy.address}] Debt Outstanding: {debt:0.3f} {symbol}")
 
-            # TODO: Actually estimate gas
-            if strategy.tendTrigger(40000 * gas_price):
+            starting_gas_price = next(gas_strategy.get_gas_price())
+
+            try:
+                tend_gas_estimate = int(
+                    GAS_BUFFER * strategy.tend.estimate_gas({"from": bot})
```
banteg, line 57: If we specify `owner=bot` when instantiating contracts and set the global gas strategy `network.gas_price(gas_strategy)`, we can omit the `{"from": bot, "gas_price": gas_strategy}` arg from all the function calls.

scripts/keep.py
```diff
@@ -1,10 +1,16 @@
...
+from brownie.network.gas.strategies import GasNowScalingStrategy
 from decimal import Decimal
 from eth_utils import is_checksum_address
 import requests
 from time import sleep
 
 
+GAS_BUFFER = 1.2
+
+gas_strategy = GasNowScalingStrategy()
```
banteg, line 11: Ideally we need a custom logic to rather cancel the tx for 21k gas on gas prices spike.

banteg (approved):
Looks good, although keepers can work at a loss if the gas prices spike and transaction is replaced a few times.

I added a few remarks on how to make the code a bit cleaner.

scripts/keep.py
fubuloubu: I think it's better as explicit

scripts/keep.py
fubuloubu: have to add to brownie

PR #54, opened 2020-11-02
Sandwich Attack

Grandthrax:
A general vault sandwich attack. test added to make sure an attacker does make a profit leaching vault profits trading in and out around harvest time.

PR #96, opened 2020-11-19
Disallow deposit and withdraw in the same block

banteg:
Inspired by https://github.com/makerdao/ds-chief/pull/1

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -86,6 +86,7 @@ governance: public(address)
 guardian: public(address)
 pendingGovernance: address
 guestList: public(GuestList)
+lastDeposit: HashMap[address, uint256]
 
 struct StrategyParams:
     performanceFee: uint256  # Strategist's fee (basis points)
@@ -649,6 +650,7 @@ def deposit(_amount: uint256 = MAX_UINT256, _recipient: address = msg.sender) ->
         caller's address.
     @return The issued Vault shares.
     """
+    self.lastDeposit[msg.sender] = block.number
     assert not self.emergencyShutdown  # Deposits are locked out
 
     amount: uint256 = _amount
@@ -764,6 +766,7 @@ def withdraw(_shares: uint256 = MAX_UINT256, _recipient: address = msg.sender) -
         caller's address.
     @return The quantity of tokens redeemable for `_shares`.
     """
+    assert block.number > self.lastDeposit[msg.sender]
     shares: uint256 = _shares  # May reduce this number below
 
     # If _shares not specified, transfer full share balance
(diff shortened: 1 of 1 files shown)


contracts/Vault.vy
```diff
@@ -764,6 +768,7 @@ def withdraw(_shares: uint256 = MAX_UINT256, _recipient: address = msg.sender) -
         caller's address.
     @return The quantity of tokens redeemable for `_shares`.
     """
+    assert block.number > self.lastDeposit[msg.sender]
```
dmolina79, line 771: At first glance I like it, is this is easy to test for in ganache fork?

fubuloubu: We can create a "flash loan" mock and show that it fails

dereksilva: Yes, please.

fubuloubu (changes requested):
Add:
- [ ] "Flash Loan" mock that does `deposit`/`withdraw` in one block (no need for any other action)
- [ ] "Flash Loan" mock that does `transfer`/`withdraw` in one block (no need for any other action)
- [ ] Test case that shows "flash loan" failure under both deposit and transfer cases

Macarse:
Shouldn't we use @Grandthrax's tests in https://github.com/iearn-finance/yearn-vaults/pull/54 to verify this change?

Grandthrax:
> Shouldn't we use @Grandthrax's tests in #54 to verify this change?

Unfortunately this fix won't make that test pass as its not a flash loan attack

fubuloubu:
> Shouldn't we use @Grandthrax's tests in #54 to verify this change?

not the same, no. this PR just removes flash loans and single-block sandwich attacks

contracts/Vault.vy
fubuloubu: Suggested change from today's chat:
```suggestion
    assert block.timestamp > self.lastReport
```

We don't need `lastDeposit`.

The way this will work is that it'll disallow sandwich attacks via flash loans (or normal txn sandwiching) in the same block.

banteg:
What we really care about is disallowing withdrawals in the same block as `harvest`. I'll simplify the fix.

contracts/Vault.vy
```diff
@@ -764,6 +764,7 @@ def withdraw(_shares: uint256 = MAX_UINT256, _recipient: address = msg.sender) -
         caller's address.
     @return The quantity of tokens redeemable for `_shares`.
     """
+    assert block.timestamp != self.lastReport  # dev: harvest in progress
```
fubuloubu, line 767:
```suggestion
    assert block.timestamp != self.lastReport  # dev: harvest already happened this block
```

fubuloubu:
Closing in favor of #167

PR #98, opened 2020-11-20
fix: div by zero

Macarse:
When the vault has no funds, `pricePerShare` reverts with a zero division.
That's because the external method calls `_shareValue` which does a division by `totalSuppy`.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -834,7 +834,7 @@ def pricePerShare() -> uint256:
     @return The value of a single share.
     """
     if self.totalSupply == 0:
-        return 0
+        return 10 ** self.decimals()  # price of 1:1
     else:
         return self._shareValue(10 ** self.decimals)
 
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -691,10 +691,7 @@ def deposit(_amount: uint256 = MAX_UINT256, _recipient: address = msg.sender) ->
 def _shareValue(_shares: uint256) -> uint256:
     # Determines the current value of `_shares`.
         # NOTE: if sqrt(Vault.totalAssets()) >>> 1e39, this could potentially revert
-    if self.totalSupply == 0:
-        return 0
-    else:
-        return (_shares * (self._totalAssets())) / self.totalSupply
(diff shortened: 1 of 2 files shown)


contracts/Vault.vy
```diff
@@ -691,7 +691,10 @@ def deposit(_amount: uint256 = MAX_UINT256, _recipient: address = msg.sender) ->
 def _shareValue(_shares: uint256) -> uint256:
     # Determines the current value of `_shares`.
         # NOTE: if sqrt(Vault.totalAssets()) >>> 1e39, this could potentially revert
-    return (_shares * (self._totalAssets())) / self.totalSupply
+    if self.totalSupply == 0:
+        return 0
+    else:
+        return (_shares * (self._totalAssets())) / self.totalSupply
```
fubuloubu, line 697: Can you do this in the `external` method instead? There's no reason for it here, none of the calls that use it internally suffer from the div / 0 fault, only the external view function

fubuloubu (approved):
Can you add a description of what you did in the PR?

contracts/Vault.vy
```diff
@@ -833,7 +833,10 @@ def pricePerShare() -> uint256:
     @dev See dev note on `withdraw`.
     @return The value of a single share.
     """
-    return self._shareValue(10 ** self.decimals)
+    if self.totalSupply == 0:
+        return 0
```
fubuloubu, line 837:
```suggestion
        return 10 ** self.decimals()  # price of 1:1
```

fubuloubu (changes requested):

fubuloubu (approved):

tests/functional/vault/test_config.py
```diff
@@ -30,6 +30,7 @@ def test_vault_deployment(guardian, gov, rewards, token, Vault):
     assert vault.debtOutstanding() == 0
     assert vault.maxAvailableShares() == 0
     assert vault.totalAssets() == 0
+    assert vault.pricePerShare() == 0
```
fubuloubu, line 33:
```suggestion
    assert vault.pricePerShare() / vault.decimals() == 1.0
```

fubuloubu (approved):

PR #102, opened 2020-11-21
fix: flakiness in ci

Macarse:
Removed the test which caused the issue.


fubuloubu:
Closing in favor of #64

Issue #105, opened 2020-11-24
Fix Audit BaseStrategy v0.2.0 issues


dmolina79:
Audit Issues:
- [x] use safemath in harvestTrigger multiplication
- [ ] use modifiers for auth to avoid duplication
- [x] add more events to setter functions (L156, 169, 174, 184, 189, 382)

fubuloubu:
Fixed in #107

PR #107, opened 2020-11-24
Issue 105: fixes audit report comments for BaseStrategy.sol

dmolina79:
Change Log:
* fixes for issue #105

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -139,18 +139,6 @@ abstract contract BaseStrategy {
     // So indexers can keep track of this
     event Harvested(uint256 profit, uint256 loss, uint256 debtPayment, uint256 debtOutstanding);
 
-    event StrategistUpdated(address newStrategist);
-
-    event KeeperUpdated(address newKeeper);
-
-    event RewardsUpdated(address rewards);
-
-    event ReportDelayUpdated(uint256 delay);
-
-    event ProfitFactorUpdated(uint256 profitFactor);
-
-    event DebtThresholdUpdated(uint256 debtThreshold);
-
     // The minimum number of seconds between harvest calls
     // NOTE: Override this value with your own, or set dynamically below
     uint256 public minReportDelay = 86400; // ~ once a day
@@ -164,22 +152,6 @@ abstract contract BaseStrategy {
 
     bool public emergencyExit;
 
-    // modifiers
-    modifier onlyAuthorized() {
-        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
-        _;
-    }
-
-    modifier onlyStrategist() {
-        require(msg.sender == strategist, "!strategist");
-        _;
-    }
-
-    modifier onlyGovernance() {
-        require(msg.sender == governance(), "!authorized");
-        _;
-    }
-
     constructor(address _vault) public {
         vault = VaultAPI(_vault);
         want = IERC20(vault.token());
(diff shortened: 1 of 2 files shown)


Macarse (approved):

contracts/BaseStrategy.sol
```diff
@@ -139,6 +139,18 @@ abstract contract BaseStrategy {
...
+    event KeeperUpdated(address newKeeper);
+
+    event RewardsUpdated(address rewards);
+
+    event ReportDelayUpdated(uint256 delay);
+
+    event ProfitFactorUpdated(uint256 profitFactor);
+
+    event DebtThresholdUpdated(uint256 debtThreshold);
+
```
fubuloubu, line 153: Can reverse the wording here (`Update*`) so it is easier to grok?

contracts/BaseStrategy.sol
```diff
@@ -152,6 +164,22 @@ abstract contract BaseStrategy {
 
     bool public emergencyExit;
 
+    // modifiers
+    modifier onlyAuthorized() {
+        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
```
fubuloubu, line 169: This will be one of only 2 places that `governance()` is used, so try combining that call into here

contracts/BaseStrategy.sol
```diff
@@ -152,6 +164,22 @@ abstract contract BaseStrategy {
...
+        _;
+    }
+
+    modifier onlyStrategist() {
+        require(msg.sender == strategist, "!strategist");
+        _;
+    }
+
+    modifier onlyGovernance() {
+        require(msg.sender == governance(), "!authorized");
```
fubuloubu, line 179: Same as previous

fubuloubu (approved):
Notes:
- [ ] add a modifier for the keeper logic
- [x] make each of the changes a separate commit so we can reference them individually in audit report

dmolina79:
> Notes:
> 
> * [ ]  add a modifier for the keeper logic
@fubuloubu 
On the keeper modifier do we want to keep the code as is:

```
 if (keeper != address(0)) {
            require(msg.sender == keeper || msg.sender == strategist || msg.sender == governance(), "!authorized");
        }
```

shouldn't the if condition check be removed to ensure auth checks are performed  even if keeper is 0x0 on tend and harvest?

```
 modifier onlyKeepers() {
            require(msg.sender == keeper || msg.sender == strategist || msg.sender == governance(), "!authorized");

        _;
    }
```

contracts/BaseStrategy.sol
dmolina79: done on latest commits

contracts/BaseStrategy.sol
dmolina79: do you mean we want to replace the governance method for `vault.governance()` in the modifiers? and remove the method altogether?

fubuloubu: Yes!

fubuloubu:
> shouldn't the if condition check be removed to ensure auth checks are performed even if keeper is 0x0 on tend and harvest?

The point of the 0x0 check was to allow it to be "open" if set to 0x0, but really... we should just remove this option at this point. I don't see us ever using it given how problematic it's been.

contracts/BaseStrategy.sol
dmolina79: so i added the modifiers here 83f37c8, but still governance() is used in 4 places which i thought then its useful to keep, and also is handy for viewing governance address in view method and APIs

contracts/BaseStrategy.sol
dmolina79: see above

fubuloubu (approved):

PR #110, opened 2020-11-26
Store Vault activation timestamp

banteg:
This could be used to make a universal GuestList which rate limits starting from the Vault deployment.


fubuloubu (approved):
Add a test case in `test/functional/vault/test_config.py`

PR #114, opened 2020-11-28
fix: no need to distributeRewards if balance is 0

Macarse:


contracts/BaseStrategy.sol
```diff
@@ -308,7 +308,9 @@ abstract contract BaseStrategy {
      */
     function distributeRewards() internal virtual {
         // Transfer 100% of newly-minted shares awarded to this contract to the rewards address.
-        vault.transfer(rewards, vault.balanceOf(address(this)));
+        if (vault.balanceOf(address(this)) > 0) {
```
fubuloubu, line 311: Save some gas by copying to memory first

Macarse: done

fubuloubu (approved):

PR #115, opened 2020-12-02
feat: add permit

banteg:
Implement `permit` approval by signature.

Inspired by [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612) draft, but doesn't follow it verbatim.

Notable deviations:
- EIP-712 domain `name` is fixed as `Yearn Vault` since contract name is mutable.
- EIP-712 domain `version` is read from `apiVersion`.
- `Permit` message uses `amount` instead of `value` as argument name because `value` is reserved in Vyper.
- Signature is encoded as `Bytes[65]` of packed `r, s, v` instead of three args.
- The function returns True, this part is arguable, since it's is unclear from the spec.

References:
- [Uniswap](https://github.com/Uniswap/uniswap-v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2ERC20.sol#L81)
- [DAI](https://github.com/makerdao/dss/blob/36dab5c9fdce4fc36812e4fd8228be3beb18774d/src/dai.sol#L120)
- [USDC](https://github.com/centrehq/centre-tokens/blob/5013157edecbaf5da7fb9e3afa85992965077c88/contracts/v2/Permit.sol#L68)

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1434,7 +1434,7 @@ def permit(owner: address, spender: address, amount: uint256, expiry: uint256, s
     @param signature A valid secp256k1 signature of Permit by owner encoded as r, s, v.
     """
     assert owner != ZERO_ADDRESS  # dev: invalid owner
-    assert expiry >= block.timestamp  # dev: permit expired
+    assert expiry == 0 or expiry >= block.timestamp  # dev: permit expired
     nonce: uint256 = self.nonces[owner]
     digest: bytes32 = self.permit_digest(owner, spender, amount, nonce, expiry)
     # NOTE: signature is packed as r, s, v
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1411,6 +1411,7 @@ def permit(owner: address, spender: address, amount: uint256, expiry: uint256, s
     @param amount The amount of tokens to be spent.
     @param expiry The timestamp after which the Permit is no longer valid.
     @param signature A valid secp256k1 signature of Permit by owner encoded as r, s, v.
+    @return True, if transaction completes successfully
     """
     assert owner != ZERO_ADDRESS  # dev: invalid owner
     assert expiry == 0 or expiry >= block.timestamp  # dev: permit expired
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -143,6 +143,11 @@ managementFee: public(uint256)
 performanceFee: public(uint256)
 FEE_MAX: constant(uint256) = 10_000  # 100%, or 10k basis points
 SECS_PER_YEAR: constant(uint256) = 31_557_600  # 365.25 days
+# `nonces` track `permit` approvals with signature.
+nonces: public(HashMap[address, uint256])
+DOMAIN_SEPARATOR: public(bytes32)
+DOMAIN_TYPE_HASH: constant(bytes32) = keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)')
+PERMIT_TYPE_HASH: constant(bytes32) = keccak256("Permit(address owner,address spender,uint256 amount,uint256 nonce,uint256 expiry)")
 
 
 @external
@@ -190,6 +195,16 @@ def __init__(
     self.depositLimit = MAX_UINT256  # Start unlimited
     self.lastReport = block.timestamp
     self.activation = block.timestamp
+    # EIP-712
+    self.DOMAIN_SEPARATOR = keccak256(
+        concat(
+            DOMAIN_TYPE_HASH,
+            keccak256(convert("Yearn Vault", Bytes[64])),
+            keccak256(convert(API_VERSION, Bytes[32])),
+            convert(chain.id, bytes32),
+            convert(self, bytes32)
+        )
+    )
 
 
 @pure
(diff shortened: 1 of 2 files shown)


contracts/Vault.vy
```diff
@@ -1382,3 +1397,53 @@ def sweep(_token: address, _value: uint256 = MAX_UINT256):
...
+    @param nonce The current nonce of the owner.
+    @param expiry The timestamp after which the Permit is no longer valid.
+    @param signature A valid secp256k1 signature of Permit by owner encoded as r, s, v.
+    """
+    assert owner != ZERO_ADDRESS  # dev: invalid owner
+    assert expiry >= block.timestamp  # dev: permit expired
+    assert nonce == self.nonces[owner]  # dev: invalid nonce
+    digest: bytes32 = self.permit_digest(owner, spender, amount, nonce, expiry)
+    # NOTE: signature is packed as r, s, v
+    r: uint256 = convert(slice(signature, 0, 32), uint256)
```
dmolina79, line 1442: is it better to just expect these as args like :
https://github.com/Uniswap/uniswap-v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2ERC20.sol#L81
?

banteg: Vyper can't do `uint8`, so we can't match their signature anyway. See how it's passed in the tests to see why this approach is viable, you can pass the signature arguments from `eth_accounts` signed message as is. I haven't checked if it's the case with web3.js and ethers.js.

dmolina79: yeah web3 and ethers have split signature, but if vyper doesnt have uint8 yeah its better to do it this way

contracts/Vault.vy
```diff
@@ -190,6 +195,16 @@ def __init__(
     self.depositLimit = MAX_UINT256  # Start unlimited
     self.lastReport = block.timestamp
     self.activation = block.timestamp
+    # EIP-712
+    self.DOMAIN_SEPARATOR = keccak256(
+        concat(
+            DOMAIN_TYPE_HASH,
+            keccak256(convert("Yearn Vault", Bytes[64])),
```
fubuloubu, line 202: Not sure this even changes anything lol, since encoding is via length not size
```suggestion
            keccak256(convert("Yearn Vault", Bytes[11])),
```

contracts/Vault.vy
```diff
@@ -190,6 +195,16 @@ def __init__(
...
+            keccak256(convert(API_VERSION, Bytes[32])),
```
fubuloubu, line 203: matches original above
```suggestion
            keccak256(convert(API_VERSION, Bytes[28])),
```

contracts/Vault.vy
```diff
@@ -1382,3 +1397,52 @@ def sweep(_token: address, _value: uint256 = MAX_UINT256):
     if _value == MAX_UINT256:
         value = ERC20(_token).balanceOf(self)
     self.erc20_safe_transfer(_token, self.governance, value)
+
+
+@view
+@internal
+def permit_digest(owner: address, spender: address, amount: uint256, nonce: uint256, expiry: uint256) -> bytes32:
```
fubuloubu, line 1404: You're gonna hate me, but keeping with the theme...
```suggestion
def _permitDigest(owner: address, spender: address, amount: uint256, nonce: uint256, expiry: uint256) -> bytes32:
```

contracts/Vault.vy
```diff
@@ -1382,3 +1397,52 @@ def sweep(_token: address, _value: uint256 = MAX_UINT256):
...
+    """
+    @notice
+        Approves spender by owner's signature to expend owner's tokens.
+        See https://eips.ethereum.org/EIPS/eip-2612.
+    
+    @param owner The address which is a source of funds and has signed the Permit.
+    @param spender The address which is allowed to spend the funds.
+    @param amount The amount of tokens to be spent.
+    @param expiry The timestamp after which the Permit is no longer valid.
+    @param signature A valid secp256k1 signature of Permit by owner encoded as r, s, v.
```
fubuloubu, line 1434: Add `@returns`?

contracts/Vault.vy
```diff
@@ -1382,3 +1397,52 @@ def sweep(_token: address, _value: uint256 = MAX_UINT256):
...
+    """
+    assert owner != ZERO_ADDRESS  # dev: invalid owner
```
fubuloubu, line 1436: Not strictly necessary, since it shouldn't be possible to `ecrecover` the empty address as the owner's account

contracts/Vault.vy
fubuloubu: I actually want to update `ecrecover` to except either `Bytes[65]` vrs blob, or separate v/r/s integers. 3 integers is more efficient though.

contracts/Vault.vy
```diff
@@ -1382,3 +1397,52 @@ def sweep(_token: address, _value: uint256 = MAX_UINT256):
...
+    assert expiry >= block.timestamp  # dev: permit expired
+    nonce: uint256 = self.nonces[owner]
+    digest: bytes32 = self.permit_digest(owner, spender, amount, nonce, expiry)
+    # NOTE: signature is packed as r, s, v
+    r: uint256 = convert(slice(signature, 0, 32), uint256)
+    s: uint256 = convert(slice(signature, 32, 32), uint256)
+    v: uint256 = convert(slice(signature, 64, 1), uint256)
+    assert ecrecover(digest, v, r, s) == owner  # dev: invalid signature
+    self.allowance[owner][spender] = amount
+    self.nonces[owner] += 1
```
fubuloubu, line 1446: Saves a `SLOAD`
```suggestion
    self.nonces[owner] = nonce + 1
```

tests/functional/vault/test_permit.py
```diff
@@ -0,0 +1,50 @@
...
+                {"name": "spender", "type": "address"},
+                {"name": "amount", "type": "uint256"},
+                {"name": "nonce", "type": "uint256"},
+                {"name": "expiry", "type": "uint256"},
+            ],
+        },
+        "domain": {
+            "name": "Yearn Vault",
+            "version": vault.apiVersion(),
+            "chainId": 1,
```
fubuloubu, line 33: going to fail testing unless you use `web3.eth.chainId()`

fubuloubu (approved):

contracts/Vault.vy
banteg: We need to know `owner` before recovering because their `nonce` is a part of the signed Permit.

tests/functional/vault/test_permit.py
banteg: Broken in ganache:
```
vault.chain_id() = 1
web3.eth.chain_id = 1337
web3.net.version = 1
```

banteg: Related to this bug https://github.com/trufflesuite/ganache/issues/1643

fubuloubu (approved):

contracts/Vault.vy
fubuloubu:
```suggestion
    @param signature A valid secp256k1 signature of Permit by owner encoded as r, s, v.
    @return True, if transaction completes successfully
```

dmolina79 (approved):
LGTM!

Deepcryptodive:
Sweet. So if I understand this correctly, this allows approval via signature, for third-party contracts to deposit someone's funds into Vaults? 

Let's take the example of the DAI vault
E.g.
- A third-party contract _(App X)_ wants to use the Yearn DAI Vault in the back-end to generate yield
- User Chad wants to allow _App X_ to access the DAI in his account
- Chad clicks a button in the _App X_ UI which generates the keccak256 hash 
(with `owner`= Chad's address ; `spender` =  _App X_ address; ...) 
- Chad signs the approval hash in MetaMask
- _App X_ can move the DAI into the Yearn DAI Vault via a separate transaction whenever they wish, as long as Chad does not initiate any new transactions (aka: his nonce stays the same).
- Chad does not have to pay any tx fees for all this.

Is this correct @banteg ? Please correct me if I am wrong. It feels like I might be missing something here!

fubuloubu:
> third-party contracts to deposit someone's funds into Vaults

No, only lets gasless transfers and withdrawals happen with a pre-approved message via permit

cr-igor:
Hi, is it a final decision to change three arguments `r`, `s`, and `v` into one `signature`? 

This decision breaks smart contracts integrations (which already support EIP-2612) as it changes `permit()` function signature.
Would be grate if you consider an opportunity of having the same function signature as in EIP! Thanks!

fubuloubu:
@cr-igor there already is another departure to the function signature due to a technical shortcoming in Vyper (although, `uint8` is equivalent to `uint256` in this particular circumstance as it doesn't actually save any space in the ABI), so changing to `r`/`s`/`v` wouldn't solve that issue for you.

cr-igor:
OK, I got you!
That's sad anyway(

PR #119, opened 2020-12-06
explicity check for migration success

aliatiia:
`require()` that `prepareMigration` explicitly signals its success before `transfer()`ing `want`


fubuloubu (approved):
Okay, yeah I see why this is needed. Thanks!

Please fix tests and commit messages and we should be good to merge

fubuloubu:
p.s. for commits we follow conventional commits https://www.conventionalcommits.org/en/v1.0.0/

Issue #113, opened 2020-11-28
Add two new fields to strategy contracts to simplify checking TVL and unused strategies


dudesahn:
Create two new fields to be standard in strategy contracts. 

1. Add a boolean field that indicates whether a strategy is delegating its funds to another strategy, which would greatly simplify TVL calculations. Currently, calculations for TVL to avoid double-counting needs to be curated manually. Addition of a simple boolean field to indicate whether the strategy were delegating its funds could move all of these calculations on-chain. 

Conceptually, the calculation could work as: all added strategies -> update migrations -> filter out non-TVL -> perform estimated asset count

As a name for this boolean field, I was thinking `isDelegated`, but if that would lead to confusion with the "Delegated Vaults" then we could just call it `includeInTVL`. 

2. Add a boolean field to signify whether a strategy is active or not, `isActive`. This would essentially allow "hiding" deprecated or inactive strategies without having to actually remove them from vaults. 

If desired, this could also be automated with some kind of check on value locked in the strategy.

dudesahn:
issue moved to https://github.com/iearn-finance/yearn-protocol/issues/72

fubuloubu:
@dudesahn this was the correct location for such a request

PR #120, opened 2020-12-07
feat: add TVL calculation api

fubuloubu:
fixes: #113

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -92,6 +92,8 @@ interface VaultAPI is IERC20 {
 interface StrategyAPI {
     function apiVersion() external pure returns (string memory);
 
+    function delegatedAssets() external virtual pure returns (uint256);
+
     function name() external pure returns (string memory);
 
     function vault() external view returns (address);
@@ -148,6 +150,23 @@ abstract contract BaseStrategy {
      * @return This Strategy's name.
      */
     function name() external virtual pure returns (string memory);
+
+    /**
+     * @notice
+     *  The amount (priced in want) of the total assets managed by this strategy should not count
+     *  towards Yearn's TVL calculations.
+     * @dev
+     *  You can override this field to set it to a non-zero value if some of the assets of this
+     *  Strategy is somehow delegated inside another part of of Yearn's ecosystem e.g. another Vault.
+     *  Note that this value must be strictly less than or equal to the amount provided by
+     *  `estimatedTotalAssets()` below, as the TVL calc will be total assets minus delegated assets.
+     * @return
+     *  The amount of assets this strategy manages that should not be included in Yearn's Total Value
+     *  Locked (TVL) calculation across it's ecosystem.
+     */
+    function delegatedAssets() external virtual pure returns (uint256) {
+        return 0;
+    }
 
     VaultAPI public vault;
     address public strategist;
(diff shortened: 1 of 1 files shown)


contracts/BaseStrategy.sol
```diff
@@ -149,6 +151,21 @@ abstract contract BaseStrategy {
...
     function name() external virtual pure returns (string memory);
 
+    /**
+     * @notice
+     *  The amount (priced in want) of the total assets managed by this strategy should not count
+     *  towards Yearn's TVL calculations.
+     * @dev
+     *  You can override this field to set it to a non-zero value if some of the assets of this
+     *  Strategy is somehow delegated inside another part of of Yearn's ecosystem e.g. another Vault.
+     *  Note that this value must be strictly less than or equal to the amonut provided by
```
dudesahn, line 163: *amount

contracts/BaseStrategy.sol
```diff
@@ -149,6 +153,21 @@ abstract contract BaseStrategy {
...
+    /**
+     * @notice
+     *  The amount (priced in want) of the total assets managed by this strategy should not count
+     *  towards Yearn's TVL calculations.
+     * @dev
+     *  You can override this field to set it to a non-zero value if some of the assets of this
+     *  Strategy is somehow delegated inside another part of of Yearn's ecosystem e.g. another Vault.
+     *  Note that this value must be strictly less than or equal to the amount provided by
+     *  `estimatedTotalAssets()` below, as the TVL calc will be total assets minus delegated assets.
+     * @return True if positions should be included in Yearn's Total Value Locked (TVL) calculation.
```
dmolina79, line 165: Is this comment correct? Return value is uint

contracts/BaseStrategy.sol
```diff
@@ -149,6 +153,21 @@ abstract contract BaseStrategy {
...
 
+    /**
+     * @notice
+     *  The amount (priced in want) of the total assets managed by this strategy should not count
+     *  towards Yearn's TVL calculations.
+     * @dev
+     *  You can override this field to set it to a non-zero value if some of the assets of this
+     *  Strategy is somehow delegated inside another part of of Yearn's ecosystem e.g. another Vault.
+     *  Note that this value must be strictly less than or equal to the amount provided by
+     *  `estimatedTotalAssets()` below, as the TVL calc will be total assets minus delegated assets.
```
dmolina79, line 164: Not clear on the comment, is it correct to assume estimated total assets should be a sum of total assets in control of the strat + delegated assets? If this is the case we can make this non virtual and make the two sub method virtual? 

```suggestion
     *  EstimatedTotalAssets= 
TotalAssets() + delegatedAssets()
```

dmolina79 (approved):
Left two minor comments. Code looks good

contracts/BaseStrategy.sol
dudesahn: ah yeah I think that's probably leftover from when this was originally going to be boolean

fubuloubu: crap, missed this

contracts/BaseStrategy.sol
fubuloubu: No, `delegatedAssets()` has a particular meaning for TVL calculations only. It means that amount is delegated to another Yearn component accounted for in TVL calculations, and as such should be subtracted from `estimatedTotalAssets()` here in order to conduct the calculation.

dmolina79: Makes sense, thanks for explanation

PR #122, opened 2020-12-08
fix: explicitly check for migration success

aliatiia:
strategies implementing `prepareMigration` must explicitly signal success of migration prerequisites before the the balance is transferred to the new strategy.

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -602,13 +602,13 @@ abstract contract BaseStrategy {
      * transferring any reserve or LP tokens, CDPs, or other tokens or stores of
      * value.
      */
-    function prepareMigration(address _newStrategy) internal virtual;
+    function prepareMigration(address _newStrategy) internal virtual returns (bool _success);
 
     /**
      * @notice
      *  Transfers all `want` from this Strategy to `_newStrategy`.
      *
-     *  This may only be called by governance or the Vault.
+     *  This may only be called by governance or the Vault. The Strategy must explicitly signal success of migration prerequisites before balance is transferred to `_newStrategy`
      * @dev
      *  The new Strategy's Vault must be the same as this Strategy's Vault.
      * @param _newStrategy The Strategy to migrate to.
@@ -616,7 +616,7 @@ abstract contract BaseStrategy {
     function migrate(address _newStrategy) external {
         require(msg.sender == address(vault) || msg.sender == governance());
         require(BaseStrategy(_newStrategy).vault() == vault);
-        prepareMigration(_newStrategy);
+        require(prepareMigration(_newStrategy));
         want.transfer(_newStrategy, want.balanceOf(address(this)));
     }
 
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -608,7 +608,8 @@ abstract contract BaseStrategy {
      * @notice
      *  Transfers all `want` from this Strategy to `_newStrategy`.
      *
-     *  This may only be called by governance or the Vault. The Strategy must explicitly signal success of migration prerequisites before balance is transferred to `_newStrategy`
+     *  This may only be called by governance or the Vault. The Strategy must explicitly signal success of migration
+     *  prerequisites before balance is transferred to `_newStrategy`
      * @dev
      *  The new Strategy's Vault must be the same as this Strategy's Vault.
      * @param _newStrategy The Strategy to migrate to.
--- a/contracts/test/TestStrategy.sol
+++ b/contracts/test/TestStrategy.sol
@@ -88,8 +88,9 @@ contract TestStrategy is BaseStrategy {
         }
     }
 
-    function prepareMigration(address _newStrategy) internal override {
+    function prepareMigration(address _newStrategy) internal override returns (bool _success){
         // Nothing needed here because no additional tokens/tokenized positions for mock
+        return true;
     }
 
     function protectedTokens() internal override view returns (address[] memory) {


aliatiia:
this is actually a breaking change (should've been `fix!:`), if merged please [notify](https://github.com/orgs/iearn-finance/projects/5) the [strategists](https://github.com/iearn-finance/brownie-strategy-mix/network/members)

contracts/BaseStrategy.sol
```diff
@@ -602,21 +602,21 @@ abstract contract BaseStrategy {
...
+     *  This may only be called by governance or the Vault. The Strategy must explicitly signal success of migration prerequisites before balance is transferred to `_newStrategy`
      * @dev
      *  The new Strategy's Vault must be the same as this Strategy's Vault.
      * @param _newStrategy The Strategy to migrate to.
      */
     function migrate(address _newStrategy) external {
         require(msg.sender == address(vault) || msg.sender == governance());
         require(BaseStrategy(_newStrategy).vault() == vault);
-        prepareMigration(_newStrategy);
+        require(prepareMigration(_newStrategy));
```
Macarse, line 619: what's the worst-case scenario? 
if the migration fails, want might fail.

aliatiia: e.g. a strategy doesn't cleanly transfer all AUM to the new_strategy, but parent's `migrate()` transfers `want` anyway.
this `require` nudges strategists towards thinking about and explicitly defining what a successful migration means in their context.

Macarse: mmm as a strategist I would revert the tx if I can't cleanly transfer all AUM.
Do you think that having a `bool` return will force the behavior?

API decision for @fubuloubu

aliatiia: > Do you think that having a bool return will force the behavior?

yes otherwise it should be an obvious flag for auditors, i.e. a yolo `return true` without any preceding `assert`ions

contracts/BaseStrategy.sol
```diff
@@ -602,21 +602,21 @@ abstract contract BaseStrategy {
...
      */
-    function prepareMigration(address _newStrategy) internal virtual;
+    function prepareMigration(address _newStrategy) internal virtual returns (bool _success);
 
     /**
      * @notice
      *  Transfers all `want` from this Strategy to `_newStrategy`.
      *
-     *  This may only be called by governance or the Vault.
+     *  This may only be called by governance or the Vault. The Strategy must explicitly signal success of migration prerequisites before balance is transferred to `_newStrategy`
```
fubuloubu, line 611:
```suggestion
     *  This may only be called by governance or the Vault. The Strategy must explicitly signal success of migration
     *  prerequisites before balance is transferred to `_newStrategy`
```

fubuloubu (approved):
This PR is structurally okay.

---

However, there is a question of what the expectations around migration are, and whether allowing for failure is a safe or useful condition to have.

The intent behind migrating is for one of 3 events:
1. Strategist releases a new version of the strategy, that addresses some underlying inefficiency or bug (that is not "existential) without relying on "Emergency Shutdown" mode (which requires time to follow through on and may be unnecessary to perform)
2. The on-chain conditions that a Strategy assumes have fundamentally changed, e.g. a hack event of an external protocol or the Strategy itself, or some sort of upgrade of the external protocol, where operation of the Strategy must change in order to continue functioning as designed. MIgration may not be necessary to correct this, but it may be more efficient that working through "Emergency Shutdown" mode in certain scenarios
3. Governance identifies an "existential" issue with a strategy where "Emergency Shutdown" mode is not a practical option to de-risk a strategy, and must switch to a "simple" shutdown strategy" in order to gracefully unwind without a loss

Blocking (1) under certain conditions could make sense, depending on things like timelocks, etc. to prevent additional losses from occuring, in the scenario where governance had made a premature decision without considering the end effects. Blocking (2) may reduce options available to correct the scenario, but depending on the condition (which is impoosible to know beforehand) could be problematic. Blocking (3) seems quite dangerous, as there the alternative of attempting to shutdown quickly may cause heavy losses.

---

TL;DR: I think this may be useful as a nanny mechanism under the scenario where a migration is not advisable because of on-chain conditions, however those conditions should be well-understood and communicated by the Strategist to Governance ahead of executing a migration. The potentially reduction of options for Governance to use this method under all conditions seems to be a net loss for this PR, and I would advise that all Strategists implementing this method simply make it clear through documentation what additional considerations there might be to ensure losses are as minimal as possible.

I would also advise that the `prepareMigration` function be designed to avoid failure as much as possible under scenarios where a loss event occurs.

fubuloubu:
Closing WONTFIX due to previously described reasoning.

Thank you so much for this PR, it really got me thinking of the dangerous of improperly constraining the migration path. Migrations should always be designed to succeed, and any risks or losses that should come under consideration before executing a migration should be documented by the strategist, instead of explicitly coding those constraints in the code of the `prepareMigration` function.

aliatiia:
np, closed while i was writing the following, fwiw:

> whether allowing for failure is a safe or useful condition to have.

ya this is a good option to have, but would prefer to have it explicitly handled. 

so if the strategy is currently in the spectrum between (a) smooth migration and (b) emergency shutdown, then governance can `forceMigration`: 

```solidity
function migrate(address _newStrategy, bool forceMigration) external {
        ..
        ..
        _success = prepareMigration(_newStrategy);
       require(_success || (forceMigration && msg.sender == governance()));
       want.transfer..
}
```

the implementer of `prepareMigration` can then go:

````solidity
if !condition1 || !condition2 || .... :
    return false;
````

throwing it back to governance to decide whether to `forceMigration`

Issue #47, opened 2020-10-27
Allow deploying "clones" of release vaults


fubuloubu:
It might make sense to refactor the Vault to allow deploying shallow "forwarder-style" proxies for each released version of the Vault, to reduce the overall gas usage from deploying new copies.

fubuloubu:
This requires changing the code in the constructor to an `init`, and then ensuring that the Vault is unusable until `init` has been called (and also that `init` cannot be called twice)

PR #124, opened 2020-12-10
feat: simplified vault registry

fubuloubu:
fixes: #45 
fixes: #47 

This PR takes the concept of the "Vault Registry" and simplifies it down to the bare bones: managing an ordered list of Vault releases and deployments. From these lists, it will be possible to further specialize around this, towards the final goal of making it possible to get a snapshot of the entire Yearn Vault ecosystem from one contract address.

There are two specializations for Vaults:
1. "Experimental" or community-deployed Vaults (unendorsed by Yearn)
2. "Endorsed" vaults, explicitly managed by Yearn, and considered a part of the Yearn ecosystem

TODO:
- [x] vault deprecation story
- [x] more testing (possibly event testing as well)
- [x] fuzzing test with multiple releases/deployments
- [x] Add docstrings
- [x] update Specification
- [x] add Vault release and deployment management to Vault deployment script

### DIFF
--- a/SPECIFICATION.md
+++ b/SPECIFICATION.md
@@ -7,6 +7,7 @@
 - Vault: Only user touch-point, manages funds
 - Strategy: Complex external interactions w/ limited access to Vault
 - Strategist: original creator of strategy, is in charge of monitoring its position for adverse effects.
+- Registry: On-chain registry of deployed Vaults for tokens, indexed by API version
 
 ## Vault Specification
 
@@ -71,6 +72,18 @@ NOTE: In this mode, the Strategy defines a reversionary set of actions that seek
 1. During Emergency Exit, the Strategy cannot take new debt from the connected Vault.
 1. During Emergency Exit, the Strategy can still interact with any external system, but must be able to handle any failure(s) of each of those system(s) as well as possible.
 1. Only Governance can exit Emergency Exit Mode.
+
+## Registry Specification
+
+### Normal Operation
+
+NOTE: The Registry allows an on-chain reference for Vault deployment events, to be used with external integrations. It is also possible to reconstruct the deployment and release history via on-chain calls.
+
+NOTE: The API version of each Vault deployment for a token should increase according to [Semantic Versioning](https://semver.org/). However, this is not practically enforced.
+
+1. Only Governance can register a new release for future Vaults to use when deploying via proxy deployment.
+1. Anyone can deploy a proxy deployment Vault, using the latest release, for experimental purposes
+1. Only Governance can endorse a Vault deployment for a given token with the latest release.
 
 ## Governance Specification
 
--- a/contracts/Registry.vy
+++ b/contracts/Registry.vy
@@ -65,12 +65,23 @@ def __init__():
 
 @external
 def setGovernance(governance: address):
+    """
+    @notice Starts the 1st phase of the governance transfer.
+    @dev Throws if the caller is not current governance.
+    @param governance The next governance address
+    """
     assert msg.sender == self.governance  # dev: unauthorized
     self.pendingGovernance = governance
 
 
 @external
 def acceptGovernance():
+    """
+    @notice Completes the 2nd phase of the governance transfer.
+    @dev
+        Throws if the caller is not the pending caller.
+        Emits a `NewGovernance` event.
+    """
     assert msg.sender == self.pendingGovernance  # dev: unauthorized
     self.governance = msg.sender
     log NewGovernance(msg.sender)
@@ -79,6 +90,11 @@ def acceptGovernance():
 @view
 @external
 def latestRelease() -> String[28]:
+    """
+    @notice Returns the api version of the latest release.
+    @dev Throws if no releases are registered yet.
+    @return The api version of the latest release.
+    """
     # NOTE: Throws if there has not been a release yet
     return Vault(self.releases[self.nextRelease - 1]).apiVersion()  # dev: no release
 
@@ -86,23 +102,27 @@ def latestRelease() -> String[28]:
 @view
 @external
 def latestVault(token: address) -> address:
+    """
+    @notice Returns the latest deployed vault for the given token.
+    @dev Throws if no deployments are endorsed yet for the given token.
+    @param token The token address to find the latest deployment for.
+    @return The address of the latest deployment.
+    """
     # NOTE: Throws if there has not been a deployment yet for this token
     return self.vaults[token][self.nextDeployment[token] - 1]  # dev: no vault for token
 
 
 @internal
-def _endorseVault(token: address, vault: address):
+def _addDeployment(token: address, vault: address):
     api_version: String[28] = Vault(vault).apiVersion()
 
-
-    # Check if there is an existing deployed vault for this token, and that we are not overwriting
+    # Check if there is an existing deployment for this token at the particular api version
     # NOTE: This doesn't check for strict semver-style linearly increasing release versions
     deployment_id: uint256 = self.nextDeployment[token]  # Next id in series
     if deployment_id > 0:
         current_version: String[28] = Vault(self.vaults[token][deployment_id - 1]).apiVersion()
-        assert current_version != api_version  # dev: same version
-    # else: we are adding a new asset to the ecosystem!
-    #       (typically after a successful "experimental" vault)
+        assert current_version != api_version  # dev: same api version
+    # else: we are adding a new token to the Registry
 
     # Update the latest deployment
     self.vaults[token][deployment_id] = vault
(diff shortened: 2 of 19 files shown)


contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+nextDeployment: public(HashMap[address, uint256])
+vaults: public(HashMap[address, HashMap[uint256, address]])
+
+# 2-phase commit
+governance: public(address)
+pending_governance: address
+
+
+event NewRelease:
+    event_id: indexed(uint256)
```
banteg, line 35: I believe `event_id` is redundant, we can guarantee ordering by `(blockNumber, logIndex)` natively.

contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+nextRelease: public(uint256)
+releases: public(HashMap[uint256, address])
+
+# Token.address => len(Vault.deployments)
+nextDeployment: public(HashMap[address, uint256])
+vaults: public(HashMap[address, HashMap[uint256, address]])
+
+# 2-phase commit
+governance: public(address)
+pending_governance: address
```
banteg, line 31: Snek case among spongeBobCase.

contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+event NewVault:
+    token: indexed(address)
+    event_id: indexed(uint256)
+    vault: address
+    api_version: String[28]
+
+event NewExperimentalVault:
+    token: indexed(address)
+    vault: address
+    api_version: String[28]
```
fubuloubu, line 48: @banteg: remove `event_id`

contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+def acceptGovernance():
+    assert msg.sender == self.pending_governance  # dev: unauthorized
+    self.governance = msg.sender
+
+
+@view
+@external
+def latestRelease() -> String[28]:
+    # NOTE: Throws if there has not been a release yet
+    return Vault(self.releases[self.nextRelease - 1]).apiVersion()  # dev: No Release!
```
banteg, line 72: Style: I think we use lowercase revert messages.

contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+    token: address = Vault(vault).token()
+    nextDeployment: uint256 = self.nextDeployment[token]
+    if nextDeployment > 0:
+        current_version: String[28] = Vault(self.vaults[token][nextDeployment - 1]).apiVersion()
+        assert current_version != latest_version  # dev: cannot override current version
+    # else: we are adding a new asset to the ecosystem!
+    #       (typically after a successful "experimental" vault)
+
+    # Add to the end of the list of vaults for token
+    self._addVault(token, vault)
```
fubuloubu, line 211: Consider adding `deprecateVault()` to denote Vaults that should be shut down

contracts/Vault.vy
```diff
@@ -151,12 +149,12 @@ PERMIT_TYPE_HASH: constant(bytes32) = keccak256("Permit(address owner,address sp
 
 
 @external
-def __init__(
+def initialize(
```
fubuloubu, line 152: Create internal method to handle this logic, and then have `__init__` call internal method (while keeping this one for proxy deployment)

contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+    @return The address of the newly-deployed vault
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+
+    # NOTE: Underflow if no releases created yet (this is okay)
+    vault: address = create_forwarder_to(self.releases[self.nextRelease - 1])
+
+    name: String[64] = nameOverride
+    if name == "":
+        name = concat("Yearn ", DetailedERC20(token).name(), " Vault")
```
banteg, line 148: The naming standard postulates it should be `{name} yVault` instead of `Yearn {name} Vault`. Although I like that variant, let's sneak that in.

contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+
+@external
+def newExperimentalVault(
+    token: address,
+    governance: address = msg.sender,
+    guardian: address = msg.sender,
+    nameOverride: String[64] = "",
+    symbolOverride: String[32] = "",
+) -> address:
+    # NOTE: Anyone can call this method, as a convenience to Strategist' experiments
```
banteg, line 170: Maybe merge this with newVault, which would deploy an experimental Vault if not called by governance.

contracts/Registry.vy
```diff
@@ -0,0 +1,211 @@
...
+event NewRelease:
+    event_id: indexed(uint256)
+    template: address
+    api_version: String[28]
+
+event NewVault:
+    token: indexed(address)
+    event_id: indexed(uint256)
+    vault: address
+    api_version: String[28]
```
banteg, line 43: Add `tags`. A vault can have a string of `tags` assigned which follow some convention. Something like `test`, `live`, `dead` and so on. An enum could be a good alternative, but imo a text tag is more flexible.

contracts/Vault.vy
banteg: Blocked by https://github.com/vyperlang/vyper/issues/2251, need a new Vyper release which allows calling internal methods from constructor.

contracts/Registry.vy
fubuloubu: oopsies

contracts/Registry.vy
fubuloubu: I don't like this because it makes the call contextual. It would be bad to mix something up here

banteg (approved):

contracts/Registry.vy
```diff
@@ -49,6 +49,9 @@ event NewExperimentalVault:
     vault: address
     api_version: String[28]
 
+event NewGovernance:
+    governance: address
```
fubuloubu, line 53: Is this in the Vault too?

banteg:
I've made some changes, probably have broken the tests.

banteg:
Would it be useful to have an enumerable set of tokens which have vaults associated with them? Or events cover that?

fubuloubu:
@banteg could add a "NewToken" event

dmolina79 (approved):
I like the template pattern, will make deploys cheaper. Left only a question if we want to track experimental vault creators, by a guess we can pull from on chain owner

LGTM!

contracts/Registry.vy
```diff
@@ -0,0 +1,253 @@
...
+) -> address:
+    # NOTE: Anyone can call this method, as a convenience to Strategist' experiments
+    # NOTE: Underflow if no releases created yet (this is okay)
+    release_template: address = self.releases[self.nextRelease - 1]  # dev: no releases
+    vault: address = self._newVault(token, release_template, governance, governance, guardian, name, symbol)
+    # NOTE: Must initialize the Vault atomically with deploying it
+    Vault(vault).initialize(token, governance, rewards, name, symbol, guardian)
+
+    # NOTE: Don't add to list of endorsed vaults (hence no event there, so we emit here)
+    log NewExperimentalVault(token, vault, Vault(release_template).apiVersion())
```
dmolina79, line 220: do we want to index also the msg.sender, creator of the vault?

banteg:
> @banteg could add a "NewToken" event

If we use events, I don't think a change is warranted, you can as well recover all tokens from NewVault events

contracts/Registry.vy
fubuloubu: Vaults that are shutdown and fully migrated are considered "deprecated"

contracts/Registry.vy
fubuloubu: solved in 6216fda

sambacha:
Way to go @fubuloubu

contracts/Registry.vy
```diff
@@ -0,0 +1,254 @@
...
+vaults: public(HashMap[address, HashMap[uint256, address]])
+
+# 2-phase commit
+governance: public(address)
+pendingGovernance: address
+
+tags: public(HashMap[address, String[1000000]])
+banksy: public(HashMap[address, bool])  # could be anyone
+
+event NewRelease:
```
sambacha, line 37: per @fubuloubu \` list of Vault releases, and an event stream to track them (just like with deployments)\`

fubuloubu:
Blocked on:
- [ ] debugging a vyper bug w/ string comparisons in memory
- [ ] debugging a brownie bug where certain calls throw a decoder error and crash w/ coverage active

contracts/Registry.vy
```diff
@@ -0,0 +1,314 @@
...
+
+
+@external
+def setGovernance(governance: address):
+    """
+    @notice Starts the 1st phase of the governance transfer.
+    @dev Throws if the caller is not current governance.
+    @param governance The next governance address
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
```
sambacha, line 69: Should `msg.sender` be `guardian`?

sambacha (approved):
My only question is about the versioning for the API. Is there a regex available to validate versioning against? Also as to what should prompt a minor version bump vs. just a patch version bump, etc.

fubuloubu:
> Is there a regex available to validate versioning against?

No, Vyper does not implement regex for strings. The deployment script (75fee4a) uses `semantic-version` package to determine linear ordering

> Also as to what should prompt a minor version bump vs. just a patch version bump, etc.

The versioning guideline for the Vaults is not under consideration for this PR, but I do agree one should be developed separately.

SPECIFICATION.md
```diff
@@ -72,6 +73,18 @@ NOTE: In this mode, the Strategy defines a reversionary set of actions that seek
...
+## Registry Specification
+
+### Normal Operation
+
+NOTE: The Registry allows an on-chain reference for Vault deployment events, to be used with external integrations. It is also possible to reconstruct the deployment and release history via on-chain calls.
+
+NOTE: The API version of each Vault deployment for a token should increase according to [Semantic Versioning](https://semver.org/). However, this is not practically enforced.
+
+1. Only Governance can register a new release for future Vaults to use when deploying via proxy deployment.
+1. Anyone can deploy a proxy deployment Vault, using the latest release, for experimental purposes
```
dmolina79, line 85: sequence numbers not incremented?

dmolina79 (approved):
LGTM on latest commit, left a minor comment on SPEC about sequence numbering

SPECIFICATION.md
fubuloubu: Markdown does it automatically

PR #125, opened 2020-12-10
feat: better deploy script

Macarse:
This is the script I used to deploy the yHegic vault.
I used @banteg trick for account fetching.

### DIFF
--- a/scripts/deploy.py
+++ b/scripts/deploy.py
@@ -31,7 +31,7 @@ def main():
     token = Token.at(get_address("ERC20 Token: "))
     gov = get_address("Yearn Governance [ychad.eth]: ")
     rewards = get_address(
-        "Rewards contract (0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde): "
+        "Rewards contract [0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde]: "
     )
     name = input(f"Set description ['{token.name()} yVault']: ") or ""
     symbol = input(f"Set symbol ['yv{token.symbol()}']: ") or ""
--- a/scripts/deploy.py
+++ b/scripts/deploy.py
@@ -1,8 +1,10 @@
 from pathlib import Path
 import yaml
+import click
 
 from brownie import Token, Vault, accounts, network, web3
 from eth_utils import is_checksum_address
+
 
 PACKAGE_VERSION = yaml.safe_load(
     (Path(__file__).parent.parent / "ethpm-config.yaml").read_text()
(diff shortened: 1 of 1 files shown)


scripts/deploy.py
```diff
@@ -24,12 +26,14 @@ def get_address(msg: str) -> str:
...
 def main():
     print(f"You are using the '{network.show_active()}' network")
-    dev = accounts.load("dev")
+    dev = accounts.load(click.prompt("Account", type=click.Choice(accounts.load())))
     print(f"You are using: 'dev' [{dev.address}]")
     token = Token.at(get_address("ERC20 Token: "))
-    gov = get_address("Yearn Governance: ")
-    rewards = get_address("Rewards contract: ")
-    name = input(f"Set description ['yearn {token.name()}']: ") or ""
+    gov = get_address("Yearn Governance (0xFEB4acf3df3cDEA7399794D0869ef76A6EfAff52): ")
```
fubuloubu, line 32: It would be awesome if this could use `ychad.eth` via [ENS instead](https://web3py.readthedocs.io/en/stable/ens_overview.html)

scripts/deploy.py
```diff
@@ -24,12 +26,14 @@ def get_address(msg: str) -> str:
...
+    rewards = get_address(
+        "Rewards contract (0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde): "
+    )
+    name = input(f"Set description ['{token.name()} vault']: ") or ""
```
fubuloubu, line 36: This is technically correct, but the update was never done
```suggestion
    name = input(f"Set description ['{token.name()} yVault']: ") or ""
```

scripts/deploy.py
```diff
@@ -24,12 +26,14 @@ def get_address(msg: str) -> str:
...
     symbol = input(f"Set symbol ['y{token.symbol()}']: ") or ""
```
fubuloubu, line 37: ditto 
```suggestion
    symbol = input(f"Set symbol ['yv{token.symbol()}']: ") or ""
```

scripts/deploy.py
```diff
@@ -24,12 +26,14 @@ def get_address(msg: str) -> str:
...
+    dev = accounts.load(click.prompt("Account", type=click.Choice(accounts.load())))
     print(f"You are using: 'dev' [{dev.address}]")
     token = Token.at(get_address("ERC20 Token: "))
-    gov = get_address("Yearn Governance: ")
-    rewards = get_address("Rewards contract: ")
-    name = input(f"Set description ['yearn {token.name()}']: ") or ""
+    gov = get_address("Yearn Governance (0xFEB4acf3df3cDEA7399794D0869ef76A6EfAff52): ")
+    rewards = get_address(
+        "Rewards contract (0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde): "
+    )
```
fubuloubu, line 35: Have a default setting for this (that uses the above address)

scripts/deploy.py
fubuloubu:
```suggestion
    gov = get_address("Yearn Governance [ychad.eth]: ")
```

Also the same default setting comment

scripts/deploy.py
Macarse: ok!

fubuloubu (approved):

PR #130, opened 2020-12-15
feat: fixes issue #129

### LINKED ISSUE #129: Avoid calling adjustPosition when there is not enough want
If the strategy doesn't have more `want` than `debtOutstanding` there is no need to call `adjustPosition`.

Macarse:
Fixes #129

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -603,7 +603,9 @@ abstract contract BaseStrategy {
         distributeRewards();
 
         // Check if free returns are left, and re-invest them
-        adjustPosition(debtOutstanding);
+        if (want.balanceOf(address(this)) > debtOutstanding) {
+            adjustPosition(debtOutstanding);
+        }
 
         emit Harvested(profit, loss, debtPayment, debtOutstanding);
     }


fubuloubu (approved):

fubuloubu:
Actually, is this really an issue?

Let's say there is outstanding debt, but it's locked away in some protocol until a time `T` in the future, which the strategy must actively make a call to now in order to withdraw at time `T`. We want the strategy to stop re-upping the lock so it can pay back the debt at time `T`, instead of never paying it back because of this condition caused a race condition where nothing ever gets unlocked because `adjustPosition` never gets called, and the signal to unlock is never made.

Macarse:
> Actually, is this really an issue?
> 
> Let's say there is outstanding debt, but it's locked away in some protocol until a time `T` in the future, which the strategy must actively make a call to now in order to withdraw at time `T`. We want the strategy to stop re-upping the lock so it can pay back the debt at time `T`, instead of never paying it back because of this condition caused a race condition where nothing ever gets unlocked because `adjustPosition` never gets called, and the signal to unlock is never made.


This is not an issue per se, it's just moving up logic is in almost every `adjustPosition` code.
Strategy shouldn't invest want if there is `debtOutstanding` so this exact same check is done inside the Strategy's `adjustPosition` code.

re your example, I am not sure how removing this check will help with the flow. If a strategy is dealing with locks in the protocol they use, they will have additional helper methods and/or flags.

fubuloubu:
My point is that the `harvest` function should not/can not be overridden, so it is not sufficient that "most" strategies are implementing these redundant checks when there are plausible ones that might find it useful to take a different tactic than what most of the others might take.

Adding `if (want.balanceOf(address(this)) < debtOutstanding) return;`in `adjustPosition` doesn't seem overly burdensome to me.

Macarse:
@fubuloubu again, I can't find an example where a strategy will do something in `adjustPosition` if `want < debtOutstanding`. There is a bigger chance someone will forget that check in the `adjustPosition` implementation.

fubuloubu:
What's the worst case scenario if this happens?

Macarse:
worst case, harvest reverts.
Common case, not a huge impact because the strategist knows how to set the debt limit.

Issue #128, opened 2020-12-14
Add Events for parameter changes in Vault


fubuloubu:
Parameter changes would be very helpful to have the history logged so they can be searched in history

hrharder:
@fubuloubu has anyone agreed to tackle this yet? if not, i'd be willing to give it a go

primarily experienced with golang and python, but have worked extensively with contracts/solidity so feel like I should be able to do this pretty easily by this weekend if its helpful

fubuloubu:
@hrharder yes, you can give it a go. The following will serve as an aide. Please add events like these in Vault.vy for governance upgrade, strategy parameter updates (rate limit, debt limit, migration, etc.) and vault parameter upgrades (deposit limit, guest list, etc.) Thanks!

https://github.com/iearn-finance/yearn-vaults/pull/124/files#diff-096c8e9d951cde1c28070e4ef10658a45c746777741fbefd68749a5ba3fe6f30R54-R55

hrharder:
great! 

you have a preference for one/a few types of events (e.g. `ParameterUpdate`) with fields `parameterName`, or many events? (e.g. `UpdateRateLimit`, `UpdateDebtLimit`, etc.)? Personally tend to prefer more granular events, but LMK

fubuloubu:
granular is better, some parameters are not `uint256`

hrharder:
lmk if this is the correct direction #131

hrharder:
since you linked events for governance updates, I removed those in the PR's 2nd commit -- happy to re add if you like (seeing you did mention governance specifically in an above comment)

fubuloubu:
@hrharder please keep them! I only added them to the Registry in that PR (still WIP), had not added them to the Vault yet.

Also, only have one event for when the transfer is accepted (e.g. `UpdateGovernance`), not when pending is set.

fubuloubu:
Reviewed what you had so far, looking good!

hrharder:
> @hrharder please keep them! I only added them to the Registry in that PR (still WIP), had not added them to the Vault yet.
> 
> Also, only have one event for when the transfer is accepted (e.g. `UpdateGovernance`), not when pending is set.

gotcha! no problem -- added [here](https://github.com/iearn-finance/yearn-vaults/pull/131/commits/641bdb2efce2a7a73a57a30f53a71c5e54fab3b4)

PR #131, opened 2020-12-17
feat: add event logs for vault config updates

hrharder:
fixes: #128 

Adds granular events for the following changes to a vaults parameters/configurations:

- ERC20 name set/updated
- ERC20 symbol set/updated
- Guest list set/updated
- Rewards address set/updated
- Deposit limit set/updated
- Performance fee set/updated
- Management fee set/updated
- Guardian address set/updated
- Emergency shutdown state set (not logged in initialization)
- Withdrawal queue set/updated

Are tests necessary for something this simple? Any feedback/requests on naming or indexed parameters or anything like that? First time using vyper as so appreciate any possible improvements.

Also -- to confirm, no linter for vyper? Followed pattern of two newlines after event declarations.

h/t @fubuloubu

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -364,7 +364,7 @@ def setGuestList(guestList: address):
     """
     assert msg.sender == self.governance
     self.guestList = GuestList(guestList)
-    log UpdateGuestList(self.guestList.address)
+    log UpdateGuestList(guestList)
 
 
 @external
@@ -383,7 +383,7 @@ def setRewards(rewards: address):
     """
     assert msg.sender == self.governance
     self.rewards = rewards
-    log UpdateRewards(self.rewards)
+    log UpdateRewards(rewards)
 
 
 @external
@@ -400,7 +400,7 @@ def setDepositLimit(limit: uint256):
     """
     assert msg.sender == self.governance
     self.depositLimit = limit
-    log UpdateDepositLimit(self.depositLimit)
+    log UpdateDepositLimit(limit)
 
 
 @external
@@ -414,7 +414,7 @@ def setPerformanceFee(fee: uint256):
     """
     assert msg.sender == self.governance
     self.performanceFee = fee
-    log UpdatePerformanceFee(self.performanceFee)
+    log UpdatePerformanceFee(fee)
 
 
 @external
@@ -428,7 +428,7 @@ def setManagementFee(fee: uint256):
     """
     assert msg.sender == self.governance
     self.managementFee = fee
-    log UpdateManagementFee(self.managementFee)
+    log UpdateManagementFee(fee)
 
 
 @external
@@ -442,7 +442,7 @@ def setGuardian(guardian: address):
     """
     assert msg.sender in [self.guardian, self.governance]
     self.guardian = guardian
-    log UpdateGuardian(self.guardian)
+    log UpdateGuardian(guardian)
 
 
 @external
@@ -468,7 +468,7 @@ def setEmergencyShutdown(active: bool):
     """
     assert msg.sender in [self.guardian, self.governance]
     self.emergencyShutdown = active
-    log EmergencyShutdown(self.emergencyShutdown)
+    log EmergencyShutdown(active)
 
 
 @external
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -116,6 +116,10 @@ event StrategyReported:
     debtLimit: uint256
 
 
+event UpdateGovernance:
+    governance: indexed(address) # New active governance
+
+
 event UpdateGuestList:
     guestList: address # Vault guest list address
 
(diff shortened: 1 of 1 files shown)


contracts/Vault.vy
```diff
@@ -116,6 +116,46 @@ event StrategyReported:
...
 
 
+event SetName:
+    name: String[64] # New active Vault ERC20 name
+
+
+event SetSymbol:
+    symbol: String[32] # New active Vault ERC20 symbol
+
+
```
fubuloubu, line 126: Probably don't need these, should only really be used once or twice

contracts/Vault.vy
```diff
@@ -207,6 +247,13 @@ def __init__(
             convert(self, bytes32)
         )
     )
+    log SetName(self.name)
+    log SetSymbol(self.symbol)
+    log SetRewards(self.rewards)
+    log SetGuardian(self.guardian)
+    log SetPerformanceFee(self.performanceFee)
+    log SetManagementFee(self.managementFee)
+    log SetDepositLimit(self.depositLimit)
```
fubuloubu, line 256: Use the input values/literals/constants for these, saving like ~1950 gas each or so

Also, put each one after where it's set

contracts/Vault.vy
```diff
@@ -116,6 +116,46 @@ event StrategyReported:
...
+event SetPerformanceFee:
+    performanceFee: indexed(uint256) # New active performance fee
+
+
+event SetManagementFee:
+    managementFee: indexed(uint256) # New active management fee
+
+
+event SetGuardian:
+    guardian: indexed(address) # Address of the active guardian
```
fubuloubu, line 148: No need to index these

Only index `strategy` address in `UpdateStrategy*` events

fubuloubu (changes requested):
Broadly, could use you use `Update*` instead? I know it's a bit more chars, but is more meaningful. Unless you have another suggestion. `Set*` doesn't see right to me.

contracts/Vault.vy
hrharder: cool -- didn't know this re. gas

hrharder:
@fubuloubu addressed comments in [a56619d](https://github.com/iearn-finance/yearn-vaults/pull/131/commits/a56619d372c3b6316bfe85cf787466e1b3e61d93) and added events for strategy param updates -- thank you

banteg (approved):

contracts/Vault.vy
```diff
@@ -116,6 +116,70 @@ event StrategyReported:
...
+    oldVersion: indexed(address) # Old version of the strategy to be upgraded
+    newVersion: indexed(address) # New version of the strategy
+
+
+event StrategyRevoked:
+    strategy: indexed(address) # Address of the strategy that is revoked
+
+
+event StrategyRemoved:
+    strategy: indexed(address) # Address of the strategy that is removed
```
fubuloubu, line 180:
```suggestion
event StrategyRemovedFromQueue:
    strategy: indexed(address) # Address of the strategy that is removed from the withdrawal queue
```

Should also have `StrategyAddedToQueue`?

contracts/Vault.vy
```diff
@@ -116,6 +116,70 @@ event StrategyReported:
...
+    strategy: indexed(address) # Address of the strategy for the rate limit adjustment
+    rateLimit: uint256 # The new rate limit for the strategy
+
+
+event UpdateStrategyPerformanceFee:
+    strategy: indexed(address) # Address of the strategy for the performance fee adjustment
+    performanceFee: uint256 # The new performance fee for the strategy
+
+
+event UpdateStrategy:
```
fubuloubu, line 170: Change to `StrategyMigrated`

contracts/Vault.vy
```diff
@@ -116,6 +116,70 @@ event StrategyReported:
...
+
+
+event UpdateStrategyRateLimit:
+    strategy: indexed(address) # Address of the strategy for the rate limit adjustment
+    rateLimit: uint256 # The new rate limit for the strategy
+
+
+event UpdateStrategyPerformanceFee:
+    strategy: indexed(address) # Address of the strategy for the performance fee adjustment
+    performanceFee: uint256 # The new performance fee for the strategy
```
fubuloubu, line 167: Maybe these could be `StrategyUpdate*` instead? To match the rest

contracts/Vault.vy
```diff
@@ -116,6 +116,70 @@ event StrategyReported:
     debtLimit: uint256
 
 
+event UpdateGovernance:
+    governance: indexed(address) # New active governance
```
fubuloubu, line 120: Not sure if this needs to be indexed

fubuloubu (approved):
A couple nitpicks, but overall looking good!

hrharder:
updated

fubuloubu (approved):

fubuloubu:
@hrharder good to go? lgtm

hrharder:
@fubuloubu indeed it is -- send it, at your leisure

PR #137, opened 2020-12-25
feat: migrateStrategy shouldn't reset activation time

### LINKED ISSUE #136: migrateStrategy shouldn't reset activation time
`migrateStrategy` shouldn't empty the replaced strategy params, it should do the same thing as revoke.

Macarse:
fixes #136

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1139,6 +1139,13 @@ def updateStrategyPerformanceFee(
     log StrategyUpdatePerformanceFee(strategy, performanceFee)
 
 
+@internal
+def _revokeStrategy(strategy: address):
+    self.debtLimit -= self.strategies[strategy].debtLimit
+    self.strategies[strategy].debtLimit = 0
+    log StrategyRevoked(strategy)
+
+
 @external
 def migrateStrategy(oldVersion: address, newVersion: address):
     """
@@ -1162,7 +1169,7 @@ def migrateStrategy(oldVersion: address, newVersion: address):
     assert self.strategies[newVersion].activation == 0
 
     strategy: StrategyParams = self.strategies[oldVersion]
-    self.strategies[oldVersion] = empty(StrategyParams)
+    self._revokeStrategy(oldVersion)
     self.strategies[newVersion] = strategy
 
     Strategy(oldVersion).migrate(newVersion)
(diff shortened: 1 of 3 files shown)


tests/functional/vault/test_strategies.py
```diff
@@ -137,8 +137,9 @@ def test_migrateStrategy(gov, vault, strategy, rando, TestStrategy):
     new_strategy = gov.deploy(TestStrategy, vault)
     vault.migrateStrategy(strategy, new_strategy, {"from": gov})
 
-    # Can migrate back again (but it starts out fresh)
-    vault.migrateStrategy(new_strategy, strategy, {"from": gov})
+    # Can't migrate back again
```
Macarse, line 140: since activation != 0, it can't be migrated back.

contracts/Vault.vy
```diff
@@ -1117,6 +1117,13 @@ def updateStrategyPerformanceFee(
     log StrategyUpdatePerformanceFee(strategy, performanceFee)
 
 
+@internal
+def _revokeStrategy(strategy: address):
```
Macarse, line 1121: I couldn't find a way to call an external method so I made the implementation `internal`

contracts/Vault.vy
```diff
@@ -1140,7 +1147,7 @@ def migrateStrategy(oldVersion: address, newVersion: address):
     assert self.strategies[newVersion].activation == 0
 
     strategy: StrategyParams = self.strategies[oldVersion]
-    self.strategies[oldVersion] = empty(StrategyParams)
+    self._revokeStrategy(oldVersion)
     self.strategies[newVersion] = strategy
```
fubuloubu, line 1151: This is probably not the best way to handle this, probably want to individually set the values, namely the parameters (don't copy the gains/losses) and total debt

Macarse: your point is that we would consider it a completely new strat?

fubuloubu: The past performance of the old strategy and the future performance of the new strategy should be treated separately

Macarse: Did all the changes you requested but copied `totalDebt`.

fubuloubu (approved):

Issue #82, opened 2020-11-15
Withdraw more than expected causes underflow


banteg:
If `liquidatePosition(uint256 _amountNeeded)(uint256 _amountFreed)` frees more than expected, it causes underflow down the line. This should be handled at BaseStrategy level and not cause an error. 

Use this Strategy fix for now
https://github.com/banteg/strategy-uni-lp-pickle/commit/fc8da9e0133b59dc19e26f5df79f6e2d836a7f78

Macarse:
yHegic strategies are also doing that same fix:
https://github.com/Macarse/yhegic/blob/master/contracts/strategies/StrategyHegic.sol#L102-L103

banteg:
A better fix suggested by Sam
https://github.com/banteg/strategy-uni-lp-pickle/commit/08f849a5639086b34c9dca248662ac4330a8bedb

fubuloubu:
It is intended that the implementer be aware of the amount freed, and that it should exceed neither the amount needed nor the amount the strategy currently has available to it locally. It is only intended to use this function in `withdraw` but we can discuss further of how it could be used in other functions to reduce duplicated code in strategies.

https://github.com/iearn-finance/yearn-vaults/blob/cff924f1894cca1820a588b14d341c4fa4f384c0/contracts/BaseStrategy.sol#L373-L379

Issue #40, opened 2020-10-26
Deep withdrawals with slippage can cause a socialized loss


fubuloubu:
Because withdrawals leave some underdefined behavior to strategies to implement them as "try to liquidate up to `amountNeeded` (see below)

https://github.com/iearn-finance/yearn-vaults/blob/22d0607291980bb19b5110c46440843b2d9e3938/contracts/BaseStrategy.sol#L276-L286

We end up in a scenario where large slippage could be caused by that action, but the withdrawer would not realize that loss and instead socialize it to the other depositors (see below)

https://github.com/iearn-finance/yearn-vaults/blob/22d0607291980bb19b5110c46440843b2d9e3938/contracts/Vault.vy#L432-L440

We currently mitigate this through Strategy review, but this is not ideal

ShadeUndertree:
I'd like to pick this one up.

fubuloubu:
Consider merging `liquidatePosition()` and `exitPosition()` as `exitPosition() := liquidatePosition(_debtOutstanding)`

PR #140, opened 2020-12-29
refactor: consolidate `exitPosition` into `liquidatePosition`

fubuloubu:
fixes: #40 
fixes: #82

This PR consolidates the two similar functions `exitPosition` and `liquidatePosition` together, by adding another value that `liquidatePosition` returns (e.g. "realized loss"). The realized loss is reported during `Vault.report` (with `Strategy.emergencyShutdown` active) or `Vault.withdraw` and assessed to maintain the integrity of the Vault's balance sheet during those events. Further, the realized loss during withdrawals is fully applied to the withdrawer instead of socialized to the Vault, which should help prevent any "slippage" attacks against strategies connected to the Vault from being exploitable in practice.

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -432,23 +432,17 @@ abstract contract BaseStrategy {
     function adjustPosition(uint256 _debtOutstanding) internal virtual;
 
     /**
-     * Make as much capital as possible "free" for the Vault to take. Some
-     * slippage is allowed, since when this method is called the strategist is
-     * no longer receiving their performance fee. The goal is for the Strategy
-     * to divest as quickly as possible while not suffering exorbitant losses.
-     * This function is used during emergency exit instead of
-     * `prepareReturn()`. This method returns any realized losses incurred,
-     * and should also return the amount of `want` tokens available to repay
-     * outstanding debt to the Vault.
+     * Liquidate up to `_amountNeeded` of `want` of this strategy's positions,
+     * irregardless of slippage. Any excess will be re-invested with `adjustPosition()`.
+     * This function should return the amount of `want` tokens made available by the
+     * liquidation. If there is a difference between them, `_loss` indicates whether the
+     * difference is due to a realized loss, or if there is some other sitution at play
+     * (e.g. locked funds). This function is used during emergency exit instead of
+     * `prepareReturn()` to liquidate all of the Strategy's positions back to the Vault.
+     *
+     * NOTE: The invariant `_amountFreed + _loss <= _amountNeeded` should always be maintained
      */
-    function exitPosition(uint256 _debtOutstanding)
-        internal
-        virtual
-        returns (
-            uint256 _profit,
-            uint256 _loss,
-            uint256 _debtPayment
-        );
+    function liquidatePosition(uint256 _amountNeeded) internal virtual returns (uint256 _liquidatedAmount, uint256 _loss);
 
     /**
      *  `Harvest()` calls this function after shares are created during
@@ -584,20 +578,27 @@ abstract contract BaseStrategy {
     function harvest() external onlyKeepers {
         uint256 profit = 0;
         uint256 loss = 0;
+        uint256 debtOutstanding = vault.debtOutstanding();
         uint256 debtPayment = 0;
         if (emergencyExit) {
             // Free up as much capital as possible
-            // NOTE: Don't take performance fee in this scenario
-            (profit, loss, debtPayment) = exitPosition(vault.debtOutstanding());
+            uint256 totalAssets = estimatedTotalAssets();
+            // NOTE: use the larger of total assets or debt outstanding to book losses properly
+            (debtPayment, loss) = liquidatePosition(totalAssets > debtOutstanding ? totalAssets : debtOutstanding);
+            // NOTE: take up any remainder here as profit
+            if (debtPayment > debtOutstanding) {
+                profit = debtPayment.sub(debtOutstanding);
+                debtPayment = debtOutstanding;
+            }
         } else {
             // Free up returns for Vault to pull
-            (profit, loss, debtPayment) = prepareReturn(vault.debtOutstanding());
+            (profit, loss, debtPayment) = prepareReturn(debtOutstanding);
         }
 
         // Allow Vault to take up to the "harvested" balance of this contract,
         // which is the amount it has earned since the last time it reported to
         // the Vault.
-        uint256 debtOutstanding = vault.report(profit, loss, debtPayment);
+        debtOutstanding = vault.report(profit, loss, debtPayment);
 
         // Distribute any reward shares earned by the strategy on this report
         distributeRewards();
@@ -609,25 +610,21 @@ abstract contract BaseStrategy {
     }
 
     /**
-     * Liquidate as many assets as possible to `want`, irregardless of
-     * slippage, up to `_amountNeeded`. Any excess should be re-invested
-     * here as well.
-     */
-    function liquidatePosition(uint256 _amountNeeded) internal virtual returns (uint256 _amountFreed);
-
-    /**
      * @notice
      *  Withdraws `_amountNeeded` to `vault`.
      *
      *  This may only be called by the Vault.
      * @param _amountNeeded How much `want` to withdraw.
+     * @return _loss Any realized losses
      */
-    function withdraw(uint256 _amountNeeded) external {
+    function withdraw(uint256 _amountNeeded) external returns (uint256 _loss) {
         require(msg.sender == address(vault), "!vault");
         // Liquidate as much as possible to `want`, up to `_amount`
-        uint256 amountFreed = liquidatePosition(_amountNeeded);
+        uint256 amountFreed;
+        (amountFreed, _loss) = liquidatePosition(_amountNeeded);
         // Send it directly back (NOTE: Using `msg.sender` saves some gas here)
         want.transfer(msg.sender, amountFreed);
+        // NOTE: Reinvest anything leftover on next `tend`/`harvest`
     }
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/BaseStrategy.sol
```diff
@@ -432,23 +432,14 @@ abstract contract BaseStrategy {
...
-            uint256 _profit,
-            uint256 _loss,
-            uint256 _debtPayment
-        );
+     * Liquidate as many assets as possible to `want`, irregardless of slippage,
+     * up to `_amountNeeded`. Any excess will be re-invested with `adjustPosition()`.
+     * This function should return any realized losses incurred, and should also
+     * return the amount of `want` tokens made available by the liquidation.
+     * This function is used during emergency exit instead of `prepareReturn()` to
+     * liquidate all of the Strategy's positions back to the Vault.
```
Macarse, line 440: liquidate `_amountNeeded` of ...

contracts/BaseStrategy.sol
```diff
@@ -584,20 +575,27 @@ abstract contract BaseStrategy {
...
         if (emergencyExit) {
             // Free up as much capital as possible
-            // NOTE: Don't take performance fee in this scenario
-            (profit, loss, debtPayment) = exitPosition(vault.debtOutstanding());
+            uint256 totalAssets = estimatedTotalAssets();
+            // NOTE: use the larger of total assets or debt oustanding to get profit/loss calculation correct
+            (debtPayment, loss) = liquidatePosition(debtOutstanding > totalAssets ? debtOutstanding : totalAssets);
+            // NOTE: take up any remainder here as profit
+            if (debtPayment > debtOutstanding) {
+                profit = debtPayment.sub(debtOutstanding);
```
Macarse, line 587: I don't understand why you take care of profits here and not losses? 
Why not make `liquidatePosition()` only `returns (uint256 _amountFreed)`

contracts/BaseStrategy.sol
```diff
@@ -432,23 +432,14 @@ abstract contract BaseStrategy {
...
-    function exitPosition(uint256 _debtOutstanding)
-        internal
-        virtual
-        returns (
-            uint256 _profit,
-            uint256 _loss,
-            uint256 _debtPayment
-        );
+     * Liquidate as many assets as possible to `want`, irregardless of slippage,
+     * up to `_amountNeeded`. Any excess will be re-invested with `adjustPosition()`.
```
fubuloubu, line 436:
```suggestion
     * Liquidate up to `_amountNeeded` of `want` of this strategy's positions,
     * irregardless of slippage. Any excess will be re-invested with `adjustPosition()`.
```

contracts/BaseStrategy.sol
fubuloubu: Is this clearer?
https://github.com/iearn-finance/yearn-vaults/pull/140/files#r549917601

contracts/BaseStrategy.sol
fubuloubu: The assumption is that `liquidatePoisition` should be used to make the Vault whole as quickly as possible, up to `_amountNeeded`. Here we are sending in the maximum amount of assets possible for the Strategy to liquidate.

If there is anything left after paying back the debt, the rest is booked as profits.

Perhaps this is over complicated though? My intention was to use this function in both the `withdraw` and `emergencyShutdown` scenario, because under that scenario we want to withdraw the assets as quickly as possible, slippage be damned.

contracts/BaseStrategy.sol
Macarse: yep!

contracts/BaseStrategy.sol
Macarse: merge

contracts/BaseStrategy.sol
Macarse: my point is if you have a liquidate position that just returns the amount freed, in `harvest()` you can do:

```
if (debtPayment > debtOutstanding) {
  // profit
} else if (debtPayment < debtOutstanding) {
  // loss
}

Comment (fubuloubu) [reply contracts/BaseStrategy.sol:587]: I think I see your point. Let me see how it looks

File: contracts/BaseStrategy.sol [hunk 4]
```diff
@@ -584,20 +577,28 @@ abstract contract BaseStrategy {
...
         uint256 profit = 0;
         uint256 loss = 0;
+        uint256 debtOutstanding = vault.debtOutstanding();
         uint256 debtPayment = 0;
         if (emergencyExit) {
             // Free up as much capital as possible
-            // NOTE: Don't take performance fee in this scenario
-            (profit, loss, debtPayment) = exitPosition(vault.debtOutstanding());
+            uint256 totalAssets = estimatedTotalAssets();
+            debtPayment = liquidatePosition(totalAssets);
```
Macarse, line 585: do you mind using a different variable to make it clear? something like `liquidatedAmount`?

fubuloubu:
Consider adding an extra default parameter to `Vault.withdraw` that allows the withdrawer to specify a maximum amount of slippage loss that is acceptable (and revert above that)

fubuloubu:
Note: add back loss reporting in `liquidatePosition` for corner case when strategy has "locked" funds

contracts/BaseStrategy.sol
```diff
@@ -432,23 +432,17 @@ abstract contract BaseStrategy {
...
-        );
+     * Liquidate up to `_amountNeeded` of `want` of this strategy's positions,
+     * irregardless of slippage. Any excess will be re-invested with `adjustPosition()`.
+     * This function should return the amount of `want` tokens made available by the
+     * liquidation. If there is a difference between them, `_loss` indicates whether the
+     * difference is due to a realized loss, or if there is some other sitution at play
+     * (e.g. locked funds). This function is used during emergency exit instead of
+     * `prepareReturn()` to liquidate all of the Strategy's positions back to the Vault.
+     *
+     * NOTE: The invariant `_amountFreed <= _amountNeeded` should always be maintained
```
fubuloubu, line 443:
```suggestion
     * NOTE: The invariant `_amountFreed + _loss <= _amountNeeded` should always be maintained
```

Macarse (approved):

contracts/BaseStrategy.sol
```diff
@@ -608,26 +609,23 @@ abstract contract BaseStrategy {
...
+    function withdraw(uint256 _amountNeeded) external returns (uint256 loss) {
         require(msg.sender == address(vault), "!vault");
         // Liquidate as much as possible to `want`, up to `_amount`
-        uint256 amountFreed = liquidatePosition(_amountNeeded);
+        uint256 amountFreed;
+        (amountFreed, loss) = liquidatePosition(_amountNeeded);
         // Send it directly back (NOTE: Using `msg.sender` saves some gas here)
         want.transfer(msg.sender, amountFreed);
+        // Reinvest anything leftover
+        adjustPosition(vault.debtOutstanding());
```
dmolina79, line 628: should this be a concern of tend instead of withdraw to reinvest leftover?

contracts/BaseStrategy.sol
```diff
@@ -608,26 +609,23 @@ abstract contract BaseStrategy {
...
      */
-    function withdraw(uint256 _amountNeeded) external {
+    function withdraw(uint256 _amountNeeded) external returns (uint256 loss) {
         require(msg.sender == address(vault), "!vault");
         // Liquidate as much as possible to `want`, up to `_amount`
-        uint256 amountFreed = liquidatePosition(_amountNeeded);
+        uint256 amountFreed;
+        (amountFreed, loss) = liquidatePosition(_amountNeeded);
         // Send it directly back (NOTE: Using `msg.sender` saves some gas here)
         want.transfer(msg.sender, amountFreed);
```
dmolina79, line 626: should we do instead
`want.transfer(msg.sender, Min(amountNeeded, amountFreed))` 

for the transfer, in case some strategy borks the accounting?

fubuloubu: No, I would say that's a bad strategy. See
https://github.com/iearn-finance/yearn-vaults/pull/140/files#diff-517cca2fbf046cbfefd26edb1fec816e5b2a5571fd99927dc81d6b4be60cd2a7R443

fubuloubu: I guess I could add an assertion for this though...

contracts/BaseStrategy.sol
fubuloubu: good point, this could also be abused as an attack vector to force a strategy to rebalance itself, potentially into a bad position.

contracts/BaseStrategy.sol
dmolina79: yeah, i was thinking of how to mitigate if the invariant is not maintained, and i don't like adding a revert or require check also since it may lock up stuff, but i guess sending more than needed is still not the end of the world since withdraw on vault also checks for amount of shares, those "additional" funds just are transferred to the vault i guess.

fubuloubu: This is exactly right, looks like it is protected! Won't add the require statement.

dmolina79 (approved):
LGTM!

contracts/BaseStrategy.sol
```diff
@@ -432,23 +432,17 @@ abstract contract BaseStrategy {
...
-        );
+     * Liquidate up to `_amountNeeded` of `want` of this strategy's positions,
+     * irregardless of slippage. Any excess will be re-invested with `adjustPosition()`.
+     * This function should return the amount of `want` tokens made available by the
+     * liquidation. If there is a difference between them, `_loss` indicates whether the
+     * difference is due to a realized loss, or if there is some other sitution at play
+     * (e.g. locked funds). This function is used during emergency exit instead of
+     * `prepareReturn()` to liquidate all of the Strategy's positions back to the Vault.
+     *
+     * NOTE: The invariant `_amountFreed + _loss <= _amountNeeded` should always be maintained
```
mushroomsforest, line 443: the `_amountFreed` is same as `_liquidatedAmount` ?

Issue #139, opened 2020-12-29
Move to Percentage-based Debt Limits


fubuloubu:
Debt Limits should be %-based, with the invariant that only <= 100% of the Vault's total assets can be lent out to Strategies

Debt limits should adjust downwards based on realized losses in `report`, and credit available should be computed dynamically via `totalAssets` and `Strategy.debtLimit` [%]

Also consider renaming `Strategy.debtLimit` to `Strategy.creditLimit`

PR #141, opened 2020-12-30
feat: use debt ratio vs. absolute debt limit

fubuloubu:
fixes: #139

This PR transitions the Strategy debt limit from an absolute value to a ratio of the total assets of the Vault. This change introduces some dynamic behavior to the Vault, as this value is based on user deposits. This has the additional benefit of scaling up as the Vault's deposits fluctuate so governance/management does not have to manage these values as often.

This introduction should be safe as the use of `Vault.totalAssets()` to determine the absolute amount at the time of `Vault.report()` (from `Strategy.harvest()`) has the same safety conditions as `Vault.deposit()`/`Vault.withdraw()`. Additionally, `Strategy.harvestTrigger()` takes into account `Vault.creditAvailable(strategy)`, so integrations may need to watch for additional behavioral changes as this may make it more difficult to have the trigger evaluate true because of on-chain conditions that might affect it.

NOTE: The ratio is scaled to basis points. The maximum all of the connected strategies can have of the total assets in the Vault is 100%, although in practice ~90% is more advisible limit.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1178,6 +1178,8 @@ def migrateStrategy(oldVersion: address, newVersion: address):
     self._revokeStrategy(oldVersion)
     # _revokeStrategy will lower the debtRatio
     self.debtRatio += strategy.debtRatio
+    # Debt is migrated to new strategy
+    self.strategies[oldVersion].totalDebt = 0
 
     self.strategies[newVersion] = StrategyParams({
         performanceFee: strategy.performanceFee,
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -91,7 +91,7 @@ guestList: public(GuestList)
 struct StrategyParams:
     performanceFee: uint256  # Strategist's fee (basis points)
     activation: uint256  # Activation block.timestamp
-    debtLimit: uint256  # Maximum borrow amount (in BPS of total assets)
+    debtRatio: uint256  # Maximum borrow amount (in BPS of total assets)
     rateLimit: uint256  # Max increase in debt per second since last harvest
     lastReport: uint256  # block.timestamp of the last time a report occured
     totalDebt: uint256  # Total outstanding debt that Strategy has
@@ -101,7 +101,7 @@ struct StrategyParams:
 
 event StrategyAdded:
     strategy: indexed(address)
-    debtLimit: uint256  # Maximum borrow amount (in BPS of total assets)
+    debtRatio: uint256  # Maximum borrow amount (in BPS of total assets)
     rateLimit: uint256  # Increase/decrease per block
     performanceFee: uint256  # Strategist's fee (basis points)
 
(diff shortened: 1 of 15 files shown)


contracts/Vault.vy
```diff
@@ -1308,20 +1310,23 @@ def _creditAvailable(strategy: address) -> uint256:
     if self.emergencyShutdown:
         return 0
 
-    strategy_debtLimit: uint256 = self.strategies[strategy].debtLimit
-    strategyreceivertalDebt: uint256 = self.strategies[strategy].totalDebt
+    vault_totalAssets: uint256 = self._totalAssets()
+    vault_debtLimit: uint256 = self.debtRatio * vault_totalAssets / MAX_BPS
+    vault_totalDebt: uint256 = self.totalDebt
+    strategy_debtLimit: uint256 = self.strategies[strategy].debtRatio * vault_totalAssets / MAX_BPS
```
dmolina79, line 1316: Small suggestion: 
checking line 1282 and this, it may be useful to also have the calculation refactored into a helper
`_strategyDebtLimit(strategy: address) -> uint256` 

saves a few lines only

contracts/Vault.vy
```diff
@@ -1060,21 +1060,22 @@ def addStrategy(
     assert strategy != ZERO_ADDRESS
 
     assert msg.sender == self.governance
+    assert self.debtRatio + debtRatio <= MAX_BPS
```
dmolina79, line 1063: Question: so in a live scenario where we have 2 strats with ratios 50% and 50% (i guess we don't want full capacity always in practice) to be able to add a 3rd strat with a 10% budget governance needs at least 1 transaction before to set either 1 of the inital 2 strats down to 40% or 45% 45% , is my understanding correct?

dmolina79 (approved):
Left a minor comment, question. Overall LGTM!

contracts/Vault.vy
fubuloubu: Yes

contracts/Vault.vy
fubuloubu: Sure. And it adds `debtLimit` back to Vault

fubuloubu: On second thought, would add another `SLOAD` here

contracts/Vault.vy
```diff
@@ -695,7 +695,7 @@ def permit(owner: address, spender: address, amount: uint256, expiry: uint256, s
 
 @view
 @internal
-def receivertalAssets() -> uint256:
+def _totalAssets() -> uint256:
```
fubuloubu, line 698: I screwed this up previous lol

banteg:
Would it be useful if the debt ratio was working allocations?

Imagine having two strategies both having 5000 (50%) debtRatio. Adding a third would require adjusting the ratios of all the existing ones first. What if we could just add it at 5000 and have the ratio computed as 5000/15000 instead, with two other strategies squeezed to 33% allocation.

Grandthrax (approved):
looks good to me

banteg (approved):

PR #143, opened 2020-12-30
feat: protect against excessive losses

fubuloubu:
requires: #141

This small change allows the user to specify the maximum amount of losses they are willing to sustain during a withdrawal (see #140). The default is 0%, so any realized loss will cause an abort

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -447,10 +447,10 @@ def setPerformanceFee(fee: uint256):
         Should set this value below the maximum strategist performance fee.
 
         This may only be called by governance.
-    @param fee The new performance fee to use. 
+    @param fee The new performance fee to use.
     """
     assert msg.sender == self.governance
-    assert fee <= MAX_BPS 
+    assert fee <= MAX_BPS
     self.performanceFee = fee
     log UpdatePerformanceFee(fee)
 
@@ -465,7 +465,7 @@ def setManagementFee(fee: uint256):
     @param fee The new management fee to use.
     """
     assert msg.sender == self.governance
-    assert fee <= MAX_BPS 
+    assert fee <= MAX_BPS
     self.managementFee = fee
     log UpdateManagementFee(fee)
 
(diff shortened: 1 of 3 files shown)


contracts/Vault.vy
```diff
@@ -956,8 +963,14 @@ def withdraw(_shares: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
 
     if value > self.token.balanceOf(self):
         # We need to go get some from our strategies in the withdrawal queue
-        # NOTE: This performs forced withdrawals from each Strategy. There is
-        #       a 0.5% withdrawal fee assessed on each forced withdrawal (<= 0.5% total)
```
dmolina79, line 960: is the fee no longer considered? if maxLoss is 0 should it account for possible fees

fubuloubu: Yeah, there is no fee. That's one of the audit comments. We forgot this little piece

contracts/test/TestStrategy.sol
```diff
@@ -59,11 +59,17 @@ contract TestStrategy is BaseStrategy {
     }
 
     function liquidatePosition(uint256 _amountNeeded) internal override returns (uint256 _liquidatedAmount, uint256 _loss) {
+        uint256 totalDebt = vault.strategies(address(this)).totalDebt;
```
dmolina79, line 62: will need a rebase from latest master

contracts/Vault.vy
```diff
@@ -849,7 +849,11 @@ def maxAvailableShares() -> uint256:
 
 
 @external
-def withdraw(_shares: uint256 = MAX_UINT256, recipient: address = msg.sender) -> uint256:
+def withdraw(
+    maxShares: uint256 = MAX_UINT256,
+    recipient: address = msg.sender,
+    maxLoss: uint256 = 0,  # BPS
```
Grandthrax, line 855: a default of zero means that if there is a dust loss from rounding this will revert. Suggest setting it to 1. (0.01%)

Grandthrax (commented):
suggest setting a >0 max loss default to account for dust

contracts/Vault.vy
fubuloubu:
```suggestion
    maxLoss: uint256 = 1,  # 0.01% [BPS]
```

dmolina79 (approved):
LGTM!

PR #146, opened 2020-12-31
feat: vault deposit limit zero on initialization

### LINKED ISSUE #121: Set Deposit Limit to 0 on deployment
Currently, the deposits are unlimited on deployment. This is potentially problematic, as we shouldn't allow Vaults to be immediately available during the setup/configuration phase. Due to how deployments are handled, it would be safer to have to set this manually in order to "open up" the Vault for deposits once deployed to mainnet, which can use multicall in order to set all the parameters needed at once, including Strategy approvals, Guest List, etc.

hrharder:
fixes #121 

- Remove set deposit limit to `2 ** 256 - 1` in `initialize`
- By default, deposit limit is `0` on construction (must call `setDepositLimit` afterwards to enable deposits)
- Removes `UpdateDepositLimit` event in `initialize`
- Updates all necessary tests & documentation

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -241,7 +241,8 @@ def initialize(
         deployed.
         The performance fee is set to 10% of yield, per Strategy.
         The management fee is set to 2%, per year.
-        There is no initial deposit limit.
+        There initial deposit limit is set to 0 (deposits disabled); it must be
+        updated after initialization.
     @dev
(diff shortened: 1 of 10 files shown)


hrharder:
@fubuloubu LMK what you think -- and apologies; haven't been able to get to this until now (lots going on with day job, holidays etc.)

Macarse (approved):

fubuloubu (approved):

PR #161, opened 2021-01-25
test: add reporting gains tests

steffenix:
@fubuloubu  Would that do for: `Vault.report()` w/ 0% fees & profit ?


fubuloubu (changes requested):

fubuloubu (approved):
Increase coverage of `Vault._assessFees` from 93.8% to 100%

Issue #160, opened 2021-01-24
Management fee is charged on total assets at time of harvest leading to unfairly high fees for early LPs


Grandthrax:
![image](https://user-images.githubusercontent.com/1507729/105635339-27a31980-5e5a-11eb-9eeb-6913850fcb8c.png)

Management fee is calculated during report. It takes the current total assets and calculated by taking the difference between current timestamp and the last reported time. This is grossly unfair for the initial depositors if there is a big wait time between vault creation and first harvest. 

Let's look at Hegic for example. Hegic vault [was created 15 days ago](https://etherscan.io/tx/0x6ae08b5f843c0663af648c58adbb4545d9a721e93e060369be7e165530830a45). first harvest was [3 days ago](https://etherscan.io/tx/0xc83a10347239c913fe1fbc171982e77191943b08c1e5022b15520333ce379c53). 

So hegic vault created. Over the next two weeks it sits empty. Then money ($450k) is deposited and harvest is run. That deposit is charges management fee as if it had been in the vault for two weeks.

As we can see in the tx report users are charged about 743 HEGIC on their deposits of 1_146_733 HEGIC (0.06%) even though nothing has happened yet.

Suggest that we use total assets at last report. Not current total assets.

dougstorm:
Good catch, I would suggest mitigation on new vault deployments to deposit a small amount and run harvest as close as possible to activation, would that mitigate the issue @Grandthrax ?

fubuloubu:
Potential solution: only charge mgmt on invested funds e.g. `Vault.totalDebt`

This will eliminate issues when Vaults are first stood up but deposited funds are uninvested, and ensures an incentive exists for governance to add more strategies and use up the debt ratio effectively.

PR #162, opened 2021-01-25
fix: only charge mgmt fee on invested funds

fubuloubu:
fixes: #160


Macarse (approved):
You should add a test :)

fubuloubu:
> You should add a test :)

leave that as an exercise for the reader lol

PR #166, opened 2021-01-26
Deploy script improvements

matkam:
The `newExperimentalVault(...)` function in the [current registry contract](https://etherscan.io/address/0xe15461b18ee31b7379019dc523231c57d1cbc18c#writeContract) expects `guardian` at argument index 2 instead of 3.

Also when deploying experimental vaults, default `governance` to the `dev` address to make it easier to set it up.

### DIFF
--- a/scripts/deploy.py
+++ b/scripts/deploy.py
@@ -57,7 +57,13 @@ def main():
         use_proxy = True
 
     token = Token.at(get_address("ERC20 Token"))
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):

fubuloubu:
@matkam fix your commit messages to work with [commitlint](https://commitlint.js.org/#/), and run `black` on your changes, and we should be good to go!

PR #167, opened 2021-01-27
Profit lock up that increases share price steadily over time

Grandthrax:
This solution reduces the pricePerShare by an amount equivalent to the locked up profit.

When a strategy reports a gain it is stored as lockedProfit. This amount is degraded per second so that after a certain period of time it can all be withdrawn. Default is 6 hours. 

If user withdraws on same block but after the harvest they will get a withdrawal as if harvest didn't happen. 

User deposits go in at the 'full share price'. If they deposit in the period of profit lockup they need to wait until the period is over before they get all their money back. This works like a withdrawal fee that only lasts 6 hours. 

This should stop the sandwich attack as the attacker would need to wait the full 6 hours before they can withdraw and get the profit share they are sandwiching.


contracts/Vault.vy
```diff
@@ -437,7 +449,6 @@ def setDepositLimit(limit: uint256):
     self.depositLimit = limit
     log UpdateDepositLimit(limit)
 
-
```
fubuloubu, line 440: keep the double line spacing

chriswessels:
Nice! This is a great solution to sandwich attacks without detrimenting regular users 🔥

contracts/Vault.vy
Grandthrax: fixed 05d5a59c07b731d8148f287f978b640bbd87dff8

lbertenasco:
@Grandthrax , As it is right now, if a user deposits while lockedProfit > 0, and instantly withdraws, he'll lose a % on his initial underlying due to the lockedProfit earnings being only calculated on withdrawal, right?
to clarify:
- deposit share-price => increase on harvest (or whenever underlying ends up in the strategy)
- withdraw share-price => depends on `lockedProfit, lockedProfitDegration and lastReport` [ always equal or less than deposit share-price]

Grandthrax:
> @Grandthrax , As it is right now, if a user deposits while lockedProfit > 0, and instantly withdraws, he'll lose a % on his initial underlying due to the lockedProfit earnings being only calculated on withdrawal, right?

Yes exactly. That's what I meant by this part in the initial comment:
_"User deposits go in at the 'full share price'. If they deposit in the period of profit lockup they need to wait until the period is over before they get all their money back. This works like a withdrawal fee that only lasts 6 hours."_

lbertenasco:
@Grandthrax  awesome! If there is a new harvest before the 6hs period. the user will be able to withdraw the with the full share-price of the first harvest right?
i.e. 
- user deposit
- 1st harvest for 10k on block 100 (withdraw-share-price not increased)
- 2nd harvest for 1wei on block 101 (withdraw-share-price increased by 10k but not 1 wei)
This might be a bit over the top, but it's something we'll need to take into account.
Maybe it'd be better to update (not overwrite) the `lockedProfit`

contracts/Vault.vy
```diff
@@ -1536,6 +1552,7 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     # Update reporting time
     self.strategies[msg.sender].lastReport = block.timestamp
     self.lastReport = block.timestamp
+    self.lockedProfit = gain # profit is locked and gradually released per block
```
dougstorm, line 1555: i was trying to see if there was a way to fix this from baseStrategy, but realized is going to complicate the strat code on the reporting, also is like "lying" to the vault about the actual gain, i think is the concern of the vault to keep accurate share price so your fix needs to be on the Vault

contracts/Vault.vy
```diff
@@ -813,9 +825,13 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
 @internal
 def _shareValue(shares: uint256) -> uint256:
     # Determines the current value of `shares`.
-    # NOTE: if sqrt(Vault.totalAssets()) > 1e37, this could potentially revert
-    # NOTE: using 1e3 for extra precision here, when decimals is low
-    return ((10 ** 3 * (shares * self._totalAssets())) / self.totalSupply) / 10 ** 3
+        # NOTE: if sqrt(Vault.totalAssets()) >>> 1e39, this could potentially revert
+    lockedFundsRatio: uint256 = (block.timestamp - self.lastReport) * self.lockedProfitDegration
```
dougstorm, line 829: i wonder if on line 959 we can have two internal ShareValues, the complete one for pricePerShare, and if the user lastDeposit was within a certain time frame (add blocktimestamp on deposit), we don't need to add this calc and save the user some gas, this only applies to users that deposited briefly before the window of lastReport.

More of an optimization.

dougstorm (approved):
LGTM! added test against attack and profit is greatly reduced compared to existing code

dougstorm:
Added some testing with a few notes i have:

the fix greatly reduces the profitability of sandwiching between harvest to almost neutral EV it seems

the only caveat i see with the approach that may be and issue, is lockedProfit is a global variable on vault

meaning if two strats are close to same block time reporting a gain, the last amount will overwrite the previous one, if the gain is bigger it seems to work out ok, but if second report gain its less, it sort of frees the fraction of funds from first strat to be available before degradation.

harvest time: strat1-> 1pm = 100 ETHER gain
harvest time: strat2-> 2pm = 50 ETHER gain

it frees up 50 ETHER from first strat harvest on the second one, unless im missing smth.

contracts/Vault.vy
```diff
@@ -219,7 +219,9 @@ debtRatio: public(uint256)  # Debt ratio for the Vault across all strategies (in
 totalDebt: public(uint256)  # Amount of tokens that all strategies have borrowed
 lastReport: public(uint256)  # block.timestamp of last report
 activation: public(uint256)  # block.timestamp of contract deployment
+lockedProfit: public(uint256) # how much profit is locked and cant be withdrawn
```
fubuloubu, line 222: If we add a mapping of per-strategy profit locked (or add it to the struct), we could make this an aggregated value, and the adjust the aggregated value during report (subtract the old per-strategy value, then add the new)

contracts/Vault.vy
```diff
@@ -855,11 +867,15 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
...
-    # NOTE: if sqrt(Vault.totalAssets()) > 1e37, this could potentially revert
-    # NOTE: using 1e3 for extra precision here, when decimals is low
-    return ((10 ** 3 * (shares * self._totalAssets())) / self.totalSupply) / 10 ** 3
-
-
+        # NOTE: if sqrt(Vault.totalAssets()) >>> 1e39, this could potentially revert
+    lockedFundsRatio: uint256 = (block.timestamp - self.lastReport) * self.lockedProfitDegration
+    freeFunds: uint256 = self._totalAssets() 
+    if(lockedFundsRatio < convert(1e18, uint256)):
+        freeFunds = freeFunds - (self.lockedProfit - (lockedFundsRatio * self.lockedProfit / convert(1e18, uint256)))
```
fubuloubu, line 874: Can probably turn this into a `min(freeFunds, ...)` statement

contracts/test/Token.sol
```diff
@@ -34,4 +34,9 @@ contract Token is ERC20 {
     ) public {
         super.transferFrom(sender, recipient, amount);
     }
+
+    // Public mint function. Do not use in production.
+    function mint(uint256 amount) public {
+        _mint(msg.sender, amount);
+    }
```
fubuloubu, line 41: Is this used anywhere?

tests/functional/strategy/test_misc.py
```diff
@@ -134,3 +134,55 @@ def test_set_metadataURI(gov, strategy, rando):
     assert strategy.metadataURI() == "ipfs://test2"
     with brownie.reverts():
         strategy.setMetadataURI("ipfs://fake", {"from": rando})
+
+
+def test_sandwich_attack(
+    chain, TestStrategy, web3, token, gov, vault, strategist, rando
+):
```
fubuloubu, line 141: Seems like this test show live under `functional/vault/`

tests/functional/strategy/test_misc.py
```diff
@@ -134,3 +134,55 @@ def test_set_metadataURI(gov, strategy, rando):
...
+    # strategy is returning 0.02%. Equivalent to 35.6% a year at 5 harvests a day
+    profit_to_be_returned = token.balanceOf(strategy) / 5000
+    token.transfer(strategy, profit_to_be_returned, {"from": honest_lp})
+
+    # now for the attack
+
+    # attacker sees harvest enter tx pool
+    attack_amount = balance
+
+    print(f"Attack Capital: {attack_amount}")
```
fubuloubu, line 167: No print statements please, use `assert`

tests/functional/strategy/test_withdrawal.py
```diff
@@ -20,3 +20,58 @@ def test_withdraw(chain, gov, token, vault, strategy, rando):
     # Anything over what we can liquidate is totally withdrawn
     strategy.withdraw(balance, {"from": vault.address})
     assert strategy.estimatedTotalAssets() == 0
+
+
+def test_user_withdraw(chain, gov, token, vault, strategy, rando):
```
fubuloubu, line 25: Better in `functional/vault/`?

tests/functional/strategy/test_withdrawal.py
```diff
@@ -20,3 +20,58 @@ def test_withdraw(chain, gov, token, vault, strategy, rando):
...
+    assert vault.pricePerShare() == pricePerShareBefore * 2  # profit
+    assert vault.totalAssets() == deposit * 2  # profit
+
+    vault.withdraw({"from": gov})
+
+    assert vault.totalSupply() == 0
+    assert token.balanceOf(vault) == 0  # all money withdrawn but some profit left
+
+
+def test_profit_degration(chain, gov, token, vault, strategy, rando):
```
fubuloubu, line 49: Better in `functional/vault/`?

contracts/test/Token.sol
dougstorm: i think we can remove

tests/functional/strategy/test_withdrawal.py
steffenix: Yes

contracts/Vault.vy
dougstorm: i think that ideas is good since with more strategies overwriting last Report could become an issue. Main thing is we can make sure it doesnt over complicate the change, since this is an improvement already over current code.

contracts/Vault.vy
fubuloubu: This at least makes things a little cleaner (the exact same code)
```suggestion
        freeFunds -= self.lockedProfit - (lockedFundsRatio * self.lockedProfit / convert(1e18, uint256))
```

fubuloubu: Also, why bother doing `convert(1e18, uint256)` when you can just do `10 ** 18`?

contracts/Vault.vy
fubuloubu: Turn this into an issue, and we'll loop back to it later

tests/functional/strategy/test_withdrawal.py
dougstorm: solved in latest

tests/functional/strategy/test_withdrawal.py
dougstorm: done in latest

tests/functional/strategy/test_misc.py
dougstorm: done in latest

tests/functional/strategy/test_misc.py
dougstorm: done in latest

contracts/test/Token.sol
dougstorm: removed in latest

contracts/Vault.vy
dougstorm: Done
https://github.com/iearn-finance/yearn-vaults/issues/196

dougstorm (approved):
Looks good based on latest changes suggested!

contracts/Vault.vy
steffenix: I think the 10 ** 18 should be a constant define on top

dougstorm: resolved in latest

contracts/Vault.vy
```diff
@@ -428,6 +432,15 @@ def setRewards(rewards: address):
     self.rewards = rewards
     log UpdateRewards(rewards)
 
+@external
+def setLockedProfitDegration(degration: uint256):
```
fubuloubu, line 436: Should `degration` be limited to within a range?

steffenix: It should not be bigger than the `DEGREDATION_COEFFICIENT`

dougstorm: added in 62812d6

contracts/Vault.vy
```diff
@@ -855,11 +869,15 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
 @internal
 def _shareValue(shares: uint256) -> uint256:
     # Determines the current value of `shares`.
-    # NOTE: if sqrt(Vault.totalAssets()) > 1e37, this could potentially revert
-    # NOTE: using 1e3 for extra precision here, when decimals is low
-    return ((10 ** 3 * (shares * self._totalAssets())) / self.totalSupply) / 10 ** 3
```
fubuloubu, line 860: Oops, just noticed this was dropped

contracts/Vault.vy
```diff
@@ -855,11 +869,15 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
...
+        # NOTE: if sqrt(Vault.totalAssets()) >>> 1e39, this could potentially revert
+    lockedFundsRatio: uint256 = (block.timestamp - self.lastReport) * self.lockedProfitDegration
+    freeFunds: uint256 = self._totalAssets()
 
+    if(lockedFundsRatio < DEGREDATION_COEFFICIENT):
+        freeFunds -= (self.lockedProfit - (lockedFundsRatio * self.lockedProfit / DEGREDATION_COEFFICIENT))
+    return (shares * freeFunds) / self.totalSupply
```
fubuloubu, line 878:
```suggestion
    # NOTE: using 1e3 for extra precision here, when decimals is low
    return ((10 ** 3 * (shares * freeFunds)) / self.totalSupply) / 10 ** 3
```

fubuloubu (changes requested):

contracts/Vault.vy
dougstorm: nice catch!, added in bac31e0

fubuloubu (approved):

PR #172, opened 2021-01-29
feat: use chi tokens for harvest and tend

### LINKED ISSUE #21: Add CHI support for strategies
As discussed would be useful for `harvest()` calls which should be called even with 500 gwei.

steffenix:
closes #21


steffenix:
Looking at the doc, I don't see a way to mock the contract, since tests aren't forking mainnet, I am looking for suggesting on how to get the CHI contract on `0x0000000000004946c0e9F43F4Dee607b0eF1fA1c`

fubuloubu:
You would have to reconfigure it to fork mainnet, which would make thw test suite slower.

However, I think you can use `network.connect("mainnet-fork")` to connect mid-test

steffenix:
> You would have to reconfigure it to fork mainnet, which would make thw test suite slower.
> 
> However, I think you can use `network.connect("mainnet-fork")` to connect mid-test

I have decided to make the chiToken optional and can be set if needed.

contracts/BaseStrategy.sol
```diff
@@ -225,6 +233,16 @@ abstract contract BaseStrategy {
...
     }
+    modifier useChi() {
+        if (address(chiToken) == address(0)) {
+            _;
+            return;
+        }
+        uint256 gasStart = gasleft();
+        _;
+        uint256 gasSpent = 21000 + gasStart - gasleft() + 16 * msg.data.length;
+        chiToken.freeFromUpTo(msg.sender, (gasSpent + 14154) / 41130);
```
fubuloubu, line 244: comment on some of these literal values

steffenix: ok will add.

steffenix: I can't find proper explanation from the devs about the literal values.

The formula is from https://1inch-exchange.medium.com/1inch-introduces-chi-gastoken-d0bd5bb0f92b

steffenix:
Could not reproduce the test error on my machine (tried using both test and coverage) is there some flakiness on the tests?

contracts/BaseStrategy.sol
mushroomsforest: the last value in above link is `41947` ?
```
chi.freeFromUpTo(msg.sender, (gasSpent + 14154) / 41947);
```

steffenix: So 41947 would be the amout of gas saved by each CHI burnt.
14154 my guess is the gas used by freeFromUpTo.

fubuloubu:
@steffenix yes that is CI flakiness

tests/functional/conftest.py
```diff
@@ -51,8 +51,14 @@ def keeper(accounts):
...
 @pytest.fixture
-def strategy(gov, strategist, keeper, token, vault, TestStrategy):
+def chiToken(gov, ChiToken):
+    yield gov.deploy(ChiToken)
+
+
+@pytest.fixture
+def strategy(gov, strategist, keeper, token, vault, TestStrategy, chiToken):
     strategy = strategist.deploy(TestStrategy, vault)
+    strategy.setChiToken(chiToken.address, {"from": gov})
```
fubuloubu, line 61: Remove `strategy.setChiToken()` from here and add another test with it activated to show that it works with or without this feature added

fubuloubu (changes requested):
One suggested change, and then it would be nice if you could test this on Ropsten or another testnet and point to that txn hash showing it integrates properly.

If it's only on mainnet, we can try it ourselves.

contracts/BaseStrategy.sol
mushroomsforest: make 100% sense to me man. so it is a heuristic magic value then

tests/functional/conftest.py
steffenix: > One suggested change, and then it would be nice if you could test this on Ropsten or another testnet and point to that txn hash showing it integrates properly.
> 
> If it's only on mainnet, we can try it ourselves.

So you have a deployment guide?

steffenix:
> One suggested change, and then it would be nice if you could test this on Ropsten or another testnet and point to that txn hash showing it integrates properly.
> 
> If it's only on mainnet, we can try it ourselves.

Is there a deployment guide that I can look into?

steffenix:
@fubuloubu how can I deploy the project to one of the testnets? I need some doc on how to use the deploy scripts.

fubuloubu:
> @fubuloubu how can I deploy the project to one of the testnets? I need some doc/tips on how to use the deploy scripts.

The deploy script might be specific to mainnet now, we should fix that.

To get a feel for how it workz, run:
`brownie run deploy --network mainnet-fork`

If you havent added an address yet for your accounts, do:
`brownie account add`

steffenix:
Having a hard time using the script even on mainnet-fork.

```
Running 'scripts/deploy.py::main'...
You are using the 'mainnet-fork' network
Account (1): 1
Enter the password to unlock this account:
You are using: 'dev' [0xa7D1E564611Ff3e0eB8AA4eB5d752a6c0c337918]
Vault Registry [v2.registry.ychad.eth]:
Found ENS 'v2.registry.ychad.eth' [0xE15461B18EE31b7379019Dc523231C57d1Cbc18c]
Deploy Experimental Vault [y/N]: y
ERC20 Token: 0x6b175474e89094c44da98b954eedeac495271d0f
I'm sorry, but '0x6b175474e89094c44da98b954eedeac495271d0f' is not a checksummed address or valid ENS record
```

fubuloubu:
> 0x6b175474e89094c44da98b954eedeac495271d0f

Should be `0x6B175474E89094C44Da98b954EedeAC495271d0F`

steffenix:
ok got it.

Now trying to deploy on testnet from scratch I need first to get an instance of `v2.registry.ychad.eth` is there any testnet with it available?

steffenix:
@fubuloubu working! https://kovan.etherscan.io/tx/0x1cd70db49c5aaf7f519d04eae2e05dac273044881f0c081396f80300004e8ba3

contracts/BaseStrategy.sol
fubuloubu:
```suggestion
        // NOTE: Per EIP-2028, gas cost is 16 per (non-empty) byte in calldata
        uint256 gasSpent = 21000 + gasStart - gasleft() + 16 * msg.data.length;
        // NOTE: 41947 is the estimated amt of gas refund realized per CHI redeemed
        // NOTE: 14154 is the estimated cost of the call to `freeFromUpTo`
        chiToken.freeFromUpTo(msg.sender, (gasSpent + 14154) / 41130);
```

fubuloubu:
@steffenix beautiful! add the one comment and we are good to go on this

fubuloubu (approved):

contracts/BaseStrategy.sol
```diff
@@ -228,6 +236,16 @@ abstract contract BaseStrategy {
...
         _;
     }
+    modifier useChi() {
+        if (address(chiToken) == address(0)) {
+            _;
+            return;
+        }
+        uint256 gasStart = gasleft();
+        _;
+        uint256 gasSpent = 21000 + gasStart - gasleft() + 16 * msg.data.length;
```
fubuloubu, line 246: Actually, now that I think about it, there could be issues with this math, or at least it is not clear to me if a possible overflow issue could happen.

Would this be the correct way to describe the calculation? (Should not be liable to overflow)
```suggestion
        uint256 gasSpent = (gasStart - gasleft()) + 21000 + (16 * msg.data.length);
```

contracts/BaseStrategy.sol
```diff
@@ -228,6 +236,16 @@ abstract contract BaseStrategy {
...
+        chiToken.freeFromUpTo(msg.sender, (gasSpent + 14154) / 41130);
```
fubuloubu, line 247: Is there a way to check the balance of tokens available? or does `freeFromUpTo` automatically take this into account?

dougstorm (approved):
LGTM! main assumption stands is that we trust strategist to use a valid chiToken, even so only potential malicious entry point is the `freeFromUpTo` function with those nothing with strategy state or balance that i can think of.

contracts/BaseStrategy.sol
steffenix: Yes it does  automatically take this into account the number of token that can be used.

lbertenasco (changes requested):
The PR looks good and the code works.

But I'd rather not have this as part of the `BaseStrategy` code.

I'd love to see if we can build a keep3r meta-job (uses SugarMommy to pay for it's work) that allows us to deploy contracts :)

This way we only pay with our self minted KP3R credits.

WDYT?

If not, it might also be a good idea to directly use https://github.com/emilianobonassi/gas-saver-deployer and avoid adding non-strategy related code into `BaseStrategy`

steffenix:
keep3r and using a gas token are complementary, the gas token by burning smart contracts will reduce the gas cost of the contract execution while keep3r are used to pay people that are harvesting yearn vaults.

With keeper it's not required to add gas token to harvest and tend but I can suggest adding it to keep3r jobs.

sambacha:
For what its worth, you would save even more using LGT: [matnad/liquid-gas-token](https://github.com/matnad/liquid-gas-token)

Also, my understanding is that the gas refund mechanism is supposed to be eliminated in the hardfork that includes EIP1559.

fubuloubu:
> Also, my understanding is that the gas refund mechanism is supposed to be eliminated in the hardfork that includes EIP1559.

Wonder if we should look into uGAS-like models instead

steffenix:
https://github.com/ethereum/EIPs/pull/3298

Issue #171, opened 2021-01-28
Change Rate Limit feature to Min/Max per Harvest Event


fubuloubu:
Supercedes: #11

The rate limit feature hasn't been working as expected (see #11), and leads to unpredictable behavior of `harvestTrigger` because `creditAvailable` can change it's value based on the current timestamp.

The idea behind this feature is to change the `StrategyParams` structure to have a maximum (and minimum) amount of credit necessary to perform a fund transfer to the Vault.

The minimum value would allow us to ensure Strategies like @Macarse's HEGIC lot strategies only trigger a harvest when a full lot is available to transfer to it.

Issue #11, opened 2020-10-14
Rate Limiting algorithm doesn't work under status change events


fubuloubu:
- [ ] Emergency exit from `False` to `True`gives the rate limit from the last block reported (which is earlier than when the reset happens)
- [ ] reseting debt limit on a strategy back to nonzero value (following `revokeStrategy`) uses the last block reported (which is earlier than when the reset happens)

PR #173, opened 2021-01-29
feat: change "rate limit" feature to min/max harvest debt size params

fubuloubu:
fixes: #11
fixes: #171 

This change allows the Vault to specify a minimum and maximum size to the credit increase per harvest for a given strategy. This functionality is more predictable than the previous rate limit functionality, as the values do not vary with time, and it gives a deeper control over the parameters of the debt increase.

With this change, governance should be aware that there is effectively no rate limit anymore, and should balance `StrategyParams.maxDebtIncrease` and `BaseStrategy.minHarvestPeriod` to achieve the desired rate limit functionality.


dougstorm (approved):
LGTM! ill let @Macarse also give the green flag since he is the main client of this change

contracts/BaseStrategy.sol
```diff
@@ -549,6 +573,9 @@ abstract contract BaseStrategy {
         // Should not trigger if Strategy is not activated
         if (params.activation == 0) return false;
 
+        // Should not trigger if we havebn't waited long enough since previous harvest
```
Macarse, line 576: typo `havebn't`

contracts/Vault.vy
```diff
@@ -1323,17 +1360,21 @@ def _creditAvailable(strategy: address) -> uint256:
...
-    return min(available, self.token.balanceOf(self))
-
+    available = min(available, self.token.balanceOf(self))
+
+    # Adjust by min and max borrow limits (per harvest)
+    # NOTE: min increase can be used to ensure that if a strategy has a minimum
+    #       amount of capital needed to purchase a position, it's not given capital
+    #       it can't make use of yet.
+    # NOTE: max increase is used to make sure each harvest isn't bigger than what
+    #       is authoirized. This combined with adjusting min and max periods in
```
Macarse, line 1372: `authoirized`

contracts/Vault.vy
```diff
@@ -1323,17 +1360,21 @@ def _creditAvailable(strategy: address) -> uint256:
...
+    #       `BaseStrategy` can be used to effect a "rate limit" on capital increase.
+    if available < strategy_minDebtPerHarvest:
+        return 0
+    else:
+        return min(available, strategy_maxDebtPerHarvest)
```
Macarse, line 1377: this means that if I create a hegic vault, users deposit 888_000 * 10. I would need to do 10 harvests to fill a strategy.

Macarse (changes requested):
If I understand everything correctly, you are limiting credit based on a min/max but you are not doing anything with the debt ratio.

Here's a counter-example of why this is not helpful.

I start a Hegic vault with:
 - Cream strategy as withdrawal buffer
 - Eth lot buyer
 
I want a healthy debt ratio of 5% cream, 95% lot buyer.
cream would be 0, unlimited
lot buyer would be 888k, 888k

Users deposit 100k hegic to the vault.
5k would go to cream and 0 to the lot buyer.

contracts/BaseStrategy.sol
fubuloubu:
```suggestion
        // Should not trigger if we haven't waited long enough since previous harvest
```

contracts/Vault.vy
fubuloubu:
```suggestion
    #       is authorized. This combined with adjusting min and max periods in
```

contracts/Vault.vy
fubuloubu: yes

Macarse (approved):

PR #174, opened 2021-01-30
feat: pull strategist rewards rather than push

Grandthrax:
This change sets approval to allow strategists to withdraw vault LP tokens from the vault using transferFrom rather than having them distributed on every harvest. 

This saves gas by not distributing every harvest. 

It also solves a problem with the [strategist sharer contract](https://github.com/Grandthrax/Sharer/blob/master/contracts/SharerV2.sol). We cannot tell which contract sent the rewards and so cannot differentiate between rewards from different strategies on the same vault.

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -280,12 +280,11 @@ abstract contract BaseStrategy {
 
     /**
      * @notice
-     *  Used to change `rewards`. Any distributed rewards will cease flowing
-     *  to the old address and begin flowing to this address once the change
-     *  is in effect.
+     *  Used to change `rewards`. EOA or smart contract which has the permission
+     *  to pull rewards from the vault.
      *
      *  This may only be called by the strategist.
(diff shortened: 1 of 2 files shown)


contracts/BaseStrategy.sol
```diff
@@ -467,11 +470,6 @@ abstract contract BaseStrategy {
      *   See `vault.report()` for further details.
      */
     function distributeRewards() internal virtual {
-        // Transfer 100% of newly-minted shares awarded to this contract to the rewards address.
-        uint256 balance = vault.balanceOf(address(this));
-        if (balance > 0) {
-            vault.transfer(rewards, balance);
-        }
     }
```
fubuloubu, line 473: shoudl we remove this function entirely then?

Grandthrax: yes i think so

fubuloubu: let's do it

fubuloubu (approved):

PR #175, opened 2021-01-30
 fix: reentrancy guard test not failing correctly

dougstorm:
Issue #176 

Created PR for discussion @fubuloubu. At first glance it seems reentrancy guard is not working as expected,
can check out branch and run the test `test_withdrawal_with_reentrancy`. 

Issue:
  Withdraw transaction in `test_withdrawal_with_reentrancy` is not failing as expected.

Expectation: 
  Withdraw transaction should fail, when we setup strategy for reentrancy back to vault.
  
  
  
Screen Shot 2021-01-30 at 08 26 38

### DIFF
--- a/tests/functional/vault/test_withdrawal.py
+++ b/tests/functional/vault/test_withdrawal.py
@@ -249,12 +249,14 @@ def test_withdrawal_with_reentrancy(
     token.approve(vault, 2 ** 256 - 1, {"from": gov})
     vault.deposit(1000, {"from": gov})
 
-    # To simulate reentrancy we need strategy to have some balance
-    vault.transfer(strategy, vault.balanceOf(gov) // 2, {"from": gov})
-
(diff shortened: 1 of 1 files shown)


fubuloubu (commented):
Not sure i see why moving this line makes it work

Grandthrax:
> Not sure i see why moving this line makes it work

It's the other way round. Moving this line breaks it and we need your help to determine why. Without the move the test was not checking a proper reentry because harvest transferred out all the lp tokens. This update is how the test is meant to run. But the test fails.

Why is the reentry lock not working?

dougstorm:
Debugging to see if we are reaching the actual reentrancy part, i think something changed in BaseStrat that we are not reaching `liquidatePosition` in TestStrategy anymore

fubuloubu (approved):

PR #178, opened 2021-02-01
Add minime style token balance snapshots

andy8052:
This is a solution to [issue 170](https://github.com/iearn-finance/yearn-vaults/issues/170) to add token balance tracking to the vaults. It tracks user balances with `SnapshotEvents` and the total supply when that value is changed by vault deposits or withdraws.


andy8052:
Updated look good @fubuloubu thanks for helping out here

fubuloubu (changes requested):
Merge is blocked, pending further discussion on how this change would be used

fubuloubu:
Closing in favor of #167

PR #182, opened 2021-02-02
fix: deploy script uses list

fubuloubu:
Just a quick fix to the deployment script


fubuloubu:
if someone could :+1: we can merge

Macarse (approved):

PR #183, opened 2021-02-04
feat: proxy mode support for BaseStrategy via BaseStrategyInitializable

emilianobonassi:
Reduce deployment costs if you have template strategies. Instead of inherit, delegate storage for params to minimal proxies.

e.g. https://github.com/emilianobonassi/yIdleStrategies/blob/dev/contracts/StrategyIdle.sol

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -238,6 +238,10 @@ abstract contract BaseStrategy {
         _;
     }
 
+    constructor(address _vault) public {
+        _initialize(_vault, msg.sender, msg.sender, msg.sender);
+    }
+
     /**
      * @notice
      *  Initializes the Strategy, this is called only once, when the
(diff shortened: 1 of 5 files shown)


fubuloubu:
Instead of making this modification, can you add another contract `BaseStrategyInitializable` that is defined in the currenct `BaseStrategy.sol`, extending `BaseStrategy` and just adds an `initialize` function which does the same thing as the `constructor`?

This way, you just inherit from `BaseStrategyInitializable` (instead of `BaseStrategy`) if you want this feature

fubuloubu (changes requested):
Looks good but needs a test

emilianobonassi:
Thanks @fubuloubu 

Test for sure

I will add a blocker to call `init` twice (a simple book) with the respective test as I have in the idle strats

contracts/BaseStrategy.sol
```diff
@@ -236,21 +236,30 @@ abstract contract BaseStrategy {
...
+        address _keeper
+    ) internal {
         vault = VaultAPI(_vault);
         want = IERC20(vault.token());
         want.approve(_vault, uint256(-1)); // Give Vault unlimited access (might save gas)
-        strategist = msg.sender;
-        rewards = msg.sender;
-        keeper = msg.sender;
-        vault.approve(rewards, uint256(-1)); // Allow rewards to be pulled
+        strategist = _strategist;
```
fubuloubu, line 259: Add `require(address(vault) == 0x0);` here

emilianobonassi:
TODO:
- test basestrategy via minimal proxy works as intended (use similar changes did on tests for on erc20 non compliant)
- test cannot be re-initialized

tests/functional/strategy/test_config.py
```diff
@@ -33,6 +33,13 @@ def test_strategy_deployment(strategist, vault, TestStrategy):
     assert not strategy.tendTrigger(0)
 
 
+def test_strategy_no_reinit(strategist, vault, TestStrategy):
+    strategy = strategist.deploy(TestStrategy, vault)
+
+    with brownie.reverts("Strategy already initialized"):
+        strategy.initialize(vault, strategist, strategist, strategist)
+
```
fubuloubu, line 41: add a test here deploying a new proxy-version from the already deployed version, and show re-init doesn't work there either.

emilianobonassi: I was preparing it, let's see the next commit

emilianobonassi:
Done, just looking for green light from CI 🚀

fubuloubu (approved):

fubuloubu (approved):

PR #185, opened 2021-02-05
feat: add deposit wrapper

fubuloubu:
This PR adds wrapper mix-in contracts, meant to be used to facilitate integrations with Yearn Vaults.

These wrappers enable the following use cases:
- [x] Managed migrations between deployments of vaults (using Registry)
- [x] Affliate rev-share tracking, to be able to show vault deposits Registered through our affliate program (TBD)
- [x] ~~Support for claimable reward tokens (using https://github.com/Uniswap/merkle-distributor)~~

This PR also adds the Yearn "yToken" contract, which is built to enable easier UX around balance tracking and migrations for current v2 deposits


fubuloubu:
Need to add `WEB3_INFURA_PROJECT_ID` to Secrets for ENS lookup in tests

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,154 @@
...
+abstract contract BaseWrapper {
+    using SafeMath for uint256;
+
+    ERC20 public token;
+
+    // Reduce number of external calls (SLOADs stay the same)
+    VaultAPI[] private _cachedVaults;
+
+    // v2.registry.ychad.eth
+    RegistryAPI constant registry = RegistryAPI(0xE15461B18EE31b7379019Dc523231C57d1Cbc18c);
```
dougstorm, line 19: just in case may want to make this mutable by setter instead of constant, gives us flexibility to upgrade registry as last resort, guessing is less of  pain to migrate the registry than M wrappers

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,154 @@
...
+
+    function _deposit(
+        address depositor,
+        address reciever,
+        uint256 amount,
+        bool pullFunds // If true, funds need to be pulled from `depositor` via `transferFrom`
+    ) internal returns (uint256) {
+        VaultAPI best = bestVault();
+
+        if (pullFunds) token.transferFrom(depositor, address(this), amount);
```
dougstorm, line 81: weird style is not enforce on 83 same as here

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,154 @@
...
+        uint256 amount,
+        bool withdrawFromBest // If true, also withdraw from `best`
+    ) internal returns (uint256 withdrawn) {
+        VaultAPI best = bestVault();
+
+        VaultAPI[] memory vaults = allVaults();
+        _updateVaultCache(vaults);
+
+        for (uint256 id = 0; id < vaults.length; id++) {
+            if (!withdrawFromBest && vaults[id] == best) {
```
dougstorm, line 109: im not totally sure this is what we want, but shouldn't we try to actually liquidate positions from "old" vaults first instead of latest? or do we always expect a wrapper to point to 1 single "best" vault always and not many vaults?

contracts/BaseWrapper.sol
fubuloubu: fair

contracts/yToken.sol
```diff
@@ -0,0 +1,111 @@
...
+pragma solidity ^0.6.12;
+pragma experimental ABIEncoderV2;
+
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+
+import {VaultAPI, BaseWrapper} from "./BaseWrapper.sol";
+
+contract yToken is IERC20, BaseWrapper {
```
fubuloubu, line 11: Make a version that lets you do ETH <> WETH

dougstorm: we can do it in a zap also

fubuloubu: Solved in d1ba7b0

contracts/BaseWrapper.sol
fubuloubu: Fixed in 2182967

contracts/BaseWrapper.sol
fubuloubu: Fixed in e4e2418

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    constructor(address _token) public {
+        token = ERC20(_token);
+        // v2.registry.ychad.eth
+        registry = RegistryAPI(0xE15461B18EE31b7379019Dc523231C57d1Cbc18c);
+    }
+
+    function _setRegistry(address _registry) internal {
+        // In case you want to override the registry instead of re-deploying
+        registry = RegistryAPI(_registry);
```
dougstorm, line 28: is my understanding correct that the affiliate is the full admin party on this token (not yearn) since they have full access to swap registry potentially sending deposits to a vault not controlled by yearn?

Or should yearn have some role within the access for these affiliateToken?

fubuloubu: My position was that the affiliate has full control to do whatever they want, using data (managed by us) from our Registry (e.g. deployment history) and Vaults (depositor controls). I figured if they wanted to delegate additional authority to us, that's their prerogative, not ours to enforce.

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,162 @@
...
+        return
+            best
+                .deposit(amount, reciever)
+                .mul(10**3) // Add 3 extra decimals of precision
+                .mul(best.pricePerShare()) // Adjust by price of best
+                .div(10**best.decimals())
+                .div(10**3); // Add 3 extra decimals of precision
+    }
+
+    function _withdraw(
```
dougstorm, line 105: thinking if there is value in having a method `_withdrawFromVault` has the id of the vault so user can signal to withdraw from a specific vault directly? it would add an additional id param and can check if id is the best vault to not allow it i guess? just thinking its an optimize withdraw option, of course it will fail if maxShares is higher than amount

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,162 @@
...
+        address sender,
+        address receiver,
+        uint256 amount,
+        bool withdrawFromBest // If true, also withdraw from `best`
+    ) internal returns (uint256 withdrawn) {
+        VaultAPI best = bestVault();
+
+        VaultAPI[] memory vaults = allVaults();
+        _updateVaultCache(vaults);
```
dougstorm, line 114: thinking if `cached_length >0 && cache_length == num_deployments` we don't need to update cache probably, not sure how much we save by avoiding line 59?

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        for (uint256 deployment_id = cache_length; deployment_id < num_deployments; deployment_id++) {
+            vaults[deployment_id] = VaultAPI(registry.vaults(address(token), deployment_id));
+        }
+
+        return vaults;
+    }
+
+    function _updateVaultCache(VaultAPI[] memory vaults) internal {
```
dougstorm, line 57: thinking if there is a way to warm up the cache from a keeper from time to time and check if deployments length are different from cached we can update cache before withdraws and avoid users paying for the reorg in withdraw

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    // Reduce number of external calls (SLOADs stay the same)
+    VaultAPI[] private _cachedVaults;
+
+    RegistryAPI public registry;
+
+    constructor(address _token) public {
+        token = ERC20(_token);
+        // v2.registry.ychad.eth
+        registry = RegistryAPI(0xE15461B18EE31b7379019Dc523231C57d1Cbc18c);
```
dougstorm, line 23: i guess we can check if token has a latestDEployment or blow up here 

`require(registry.latestVault(address(token) != address(0))`
?

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,162 @@
...
+        uint256 alreadyDeposited = best.balanceOf(account).mul(best.pricePerShare()).div(10**best.decimals());
+        uint256 amountToMigrate = totalBalance(account).sub(alreadyDeposited);
+
+        uint256 depositLeft = best.depositLimit().sub(best.totalAssets());
+        if (amountToMigrate > depositLeft) amountToMigrate = depositLeft;
+
+        if (amountToMigrate > 0) {
+            migrated = _withdraw(account, address(this), amountToMigrate, false); // `false` = don't withdraw from `best`
+            require(migrated == amountToMigrate);
+            require(migrated == _deposit(address(this), account, migrated, false)); // `false` = don't do `transferFrom` because it's already local
```
dougstorm, line 159: after all the side effects on migrate should the invariant be true:

`totalAssetsBeforeMigrate >=totalAssetsAfterMigrate` ?

contracts/BaseWrapper.sol
fubuloubu: This is a good suggestion

contracts/BaseWrapper.sol
fubuloubu: `latestVault` reverts if there's no latest

dougstorm: yeah thinking we revert at contract deployment to not have an affiliate token with a non supported token/vault

contracts/BaseWrapper.sol
fubuloubu: Could explore later. Worst case is one release needs to be cached by the user.

contracts/BaseWrapper.sol
fubuloubu: Yeah, it should be, since the prices of the relevant deposited shares should equal out before and after.

NOTE: I had to update this check to do an abs value check of `< 10` because it caused a failure sometimes. This should be appropriate in most circumstances, but something to be aware of. Also note that the value returned by `_deposit` is approximate

contracts/BaseWrapper.sol
fubuloubu: Added in 7b60011

.github/workflows/test.yaml
```diff
@@ -1,5 +1,8 @@
 name: Test
 
+env:
+    WEB3_INFURA_PROJECT_ID: ${{ secrets.WEB3_INFURA_PROJECT_ID }}
```
steffenix, line 4: Secrets aren't shared with PR from other repos. We had to harcode the value on the mix-strategy project.

contracts/BaseWrapper.sol
fubuloubu: Related fix in 05b876b that allows limiting the migration amount. Should be a decent mitigation for the issue here..

.github/workflows/test.yaml
fubuloubu: Removed this

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,197 @@
...
+        if (vaults.length > _cachedVaults.length) {
+            _cachedVaults = vaults;
+        }
+    }
+
+    function totalVaultBalance(address account) public view returns (uint256 balance) {
+        VaultAPI[] memory vaults = allVaults();
+
+        for (uint256 id = 0; id < vaults.length; id++) {
+            balance = balance.add(vaults[id].balanceOf(account).mul(vaults[id].pricePerShare()).div(10**uint256(vaults[id].decimals())));
```
dougstorm, line 68: I was brain storming the feasibility of the following attack:
-flash loan
-deposit portion in affiliate wrapper at pricePerShare 1 in underlying vault and get 100 shares
-deposit into Vaults directly with another portion and increase pricePerShare 1.2
-withdraw from wrapper 100 shares now valued at 1.2, do i get more tokens?
-withdraw from vault back funds

Im half thinking it doesnt work, is it feasible or worth testing/validating?

dougstorm: talked with doggie, and step 3 is not feasible, since he correctly points out that cannot bump the price per share without a harvest just by deposits

dougstorm (approved):
LGTM, all my comments are resolved

poolpitako (approved):

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,197 @@
...
+
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+
+import {RegistryAPI, VaultAPI} from "./BaseStrategy.sol";
+
+abstract contract BaseWrapper {
+    using SafeMath for uint256;
+
+    ERC20 public token;
```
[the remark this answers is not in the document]
fubuloubu: It uses a few of the details functions too

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,197 @@
...
+pragma solidity ^0.6.12;
+pragma experimental ABIEncoderV2;
+
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+
+import {RegistryAPI, VaultAPI} from "./BaseStrategy.sol";
+
+abstract contract BaseWrapper {
+    using SafeMath for uint256;
```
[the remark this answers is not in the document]
fubuloubu: can add I suppose

contracts/yToken.sol
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+interface IWETH {
+    function deposit() external payable;
+
+    function withdraw(uint256 wad) external;
+}
+
+contract yWETH is yToken {
```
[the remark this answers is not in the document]
fubuloubu: No I like it here haha. They're related.

contracts/BaseStrategy.sol
```diff
@@ -18,15 +18,55 @@ struct StrategyParams {
     uint256 totalLoss;
 }
 
+interface RegistryAPI {
```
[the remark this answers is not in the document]
fubuloubu: Can move it to BaseWrapper

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,197 @@
...
+    function _withdraw(
+        address sender,
+        address receiver,
+        uint256 amount,
+        bool withdrawFromBest // If true, also withdraw from `best`
+    ) internal returns (uint256 withdrawn) {
+        VaultAPI best = bestVault();
+
+        VaultAPI[] memory vaults = allVaults();
+        _updateVaultCache(vaults);
```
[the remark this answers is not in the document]
fubuloubu: The first person to call will update the cache yes

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,197 @@
...
+
+        // `receiver` now has `withdrawn` tokens as balance
+        if (receiver != address(this)) token.transfer(receiver, withdrawn);
+    }
+
+    function _migrate(address account) internal returns (uint256) {
+        return _migrate(account, totalVaultBalance(account));
+    }
+
+    function _migrate(address account, uint256 amount) internal returns (uint256 migrated) {
```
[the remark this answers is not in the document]
fubuloubu: imagine migrating $130m at one time :grimacing:

contracts/test/AffiliateToken.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    /// @notice The EIP-712 typehash for the permit struct used by the contract
+    bytes32 public constant PERMIT_TYPEHASH = keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
+
+    /// @notice A record of states for signing / validating signatures
+    mapping(address => uint256) public nonces;
+
+    address public affiliate;
+
+    address public pendingAffiliate;
```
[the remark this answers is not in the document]
fubuloubu: There's a PR that adds this

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    function setRegistry(address _registry) external {
+        require(msg.sender == registry.governance());
+        // In case you want to override the registry instead of re-deploying
+        registry = RegistryAPI(_registry);
+    }
+
+    function bestVault() public virtual view returns (VaultAPI) {
+        return VaultAPI(registry.latestVault(address(token)));
```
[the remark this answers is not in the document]
fubuloubu: Can override this function and provide an alternative implementation if the use case needs it. The `VaultAPI` can't change after integrating this proposal, at least without the wrappers all updating.

contracts/BaseWrapper.sol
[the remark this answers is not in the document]
poolpitako: An account is a holder? when you say someone with 130m that means a contract or a whale, right?
I would prefer having an atomic migration and start with smaller position holders, once we are comfortable, we tell bigger position holders to move. 

Not concerned about about having accounts with half migrated positions?

contracts/BaseWrapper.sol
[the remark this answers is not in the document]
fubuloubu: We don't use this anywhere else actually... I think it's only needed for `yETH` wrapper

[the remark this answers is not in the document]
fubuloubu: Added to `yWETH` wrapper

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,197 @@
...
+
+        // `receiver` now has shares of `best` (worth `amount` tokens) as balance
+        return
+            best
+                .deposit(amount, reciever)
+                .mul(best.pricePerShare()) // Adjust by price of best
+                .div(10**uint256(best.decimals()));
+    }
+
+    function min(uint256 a, uint256 b) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
fubuloubu: resolved

contracts/BaseWrapper.sol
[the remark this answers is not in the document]
fubuloubu: no, if they want to migrate piecewise, that is on them

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,207 @@
...
+        if (vaults.length > _cachedVaults.length) {
+            _cachedVaults = vaults;
+        }
+    }
+
+    function totalVaultBalance(address account) public view returns (uint256 balance) {
+        VaultAPI[] memory vaults = allVaults();
+
+        for (uint256 id = 0; id < vaults.length; id++) {
+            balance = balance.add(vaults[id].balanceOf(account).mul(vaults[id].pricePerShare()).div(10**uint256(vaults[id].decimals())));
```
lbertenasco, line 82:
```suggestion
        for (uint256 i = 0; i < vaults.length; i++) {
            balance = balance.add(vaults[i].balanceOf(account).mul(vaults[i].pricePerShare()).div(10**uint256(vaults[i].decimals())));

```
why `id`? should be either `index` or `i` right?

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,207 @@
...
+        }
+    }
+
+    function _deposit(
+        address depositor,
+        address reciever,
+        uint256 amount,
+        bool pullFunds // If true, funds need to be pulled from `depositor` via `transferFrom`
+    ) internal returns (uint256) {
+        VaultAPI best = bestVault();
```
lbertenasco, line 100:
```suggestion
        VaultAPI _bestVault = bestVault();
```
better readability, underscored since it's internal/constant in this scope

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,207 @@
...
+        VaultAPI[] memory vaults = allVaults();
+
+        for (uint256 id = 0; id < vaults.length; id++) {
+            assets = assets.add(vaults[id].totalAssets().mul(vaults[id].pricePerShare()).div(10**vaults[id].decimals()));
+        }
+    }
+
+    function _deposit(
+        address depositor,
+        address reciever,
```
lbertenasco, line 96:
```suggestion
        address receiver,
```

contracts/BaseWrapper.sol
```diff
@@ -0,0 +1,207 @@
...
+                .div(10**uint256(best.decimals()));
+    }
+
+    function _withdraw(
+        address sender,
+        address receiver,
+        uint256 amount,
+        bool withdrawFromBest // If true, also withdraw from `best`
+    ) internal returns (uint256 withdrawn) {
+        VaultAPI best = bestVault();
```
lbertenasco, line 124:
```suggestion
        VaultAPI _bestVault = bestVault();
```

contracts/BaseWrapper.sol
fubuloubu: "deployment ID" is the index used here

lbertenasco (approved):

Issue #156, opened 2021-01-18
Increase Coverage of Test Suite


fubuloubu:
Analyze coverage gaps for `Vault`/`BaseStrategy`/`Registry` (using `brownie gui` after `brownie test tests/functional --coverage`) and look for ways to increase coverage of calling conditions the test suite is currently missing.

Examples:
- [x] `token.transfer()/transferFrom()` return False
- [x] `Vault.report()` w/ 0% fees & profit
- [x] `Vault.report()` w/ profit/debt payment but no tokens
- [x] `token.transfer()` w/ no return value
- [x] `Strategy.harvestTrigger()` w/ debt outstanding
- [x] `Strategy.harvestTrigger()` w/ realized loss
- [x] `Strategy.sweep()` w/ protected tokens
- [x] `Strategy.setKeeper()/setReward()/setMinReportDelay()/setProfitFactor()/setDebtThreshold()` w/ governance

steffenix:
Would like to help!

steffenix:
will help

steffenix:
I have looked at the reports for `Vault.report()` I do not see lines highlighted in orange on `_assessFees` could you elaborate the examples?

fubuloubu:
![image](https://user-images.githubusercontent.com/3859395/105633910-548b0880-5e29-11eb-81aa-f22500f34b02.png)

steffenix:
So it's about testing the leftover dust?

fubuloubu:
No, that scenario shoukd be skipped if you set the fees for governance to 0%

steffenix:
I guess you know the tests enough so that you know that a governance fee of 0 percent isn't tested because I can't understand that looking at the test coverage.

fubuloubu:
> I guess you know the tests enough so that you know that a governance fee of 0 percent isn't tested because I can't understand that looking at the test coverage.

Haha, yes. Sorry, this is the case.

fubuloubu:
Make sure you rebase your work on https://github.com/iearn-finance/yearn-vaults/pull/164 (once merged)

I noticed one of our test cases wasn't running because it was inadvertently marked as a fixture

steffenix:
What does the next one means?

`Vault.report()` w/ profit/debt payment but no tokens

fubuloubu:
> * `Vault.report()` w/ profit/debt payment but no tokens

It looks like this one actually was resolved

> * `Strategy.harvestTrigger()` w/ debt outstanding
[...]

PR #187, opened 2021-02-06
test: increase coverage for BaseStrategy.sweep()

snorkr4:
Currently, `TestStrategy` does not have any protected tokens. In order to get full test coverage of `BaseStrategy.sweep()`, this change adds a (fake) protected token address to `TestStrategy` and tests sweep on that address, checking for a revert.

It also tests that sweep reverts if given the vault shares address.

Partial progress towards #156.

### DIFF
--- a/contracts/test/TestStrategy.sol
+++ b/contracts/test/TestStrategy.sol
@@ -12,6 +12,11 @@ import {BaseStrategy, StrategyParams, VaultAPI} from "../BaseStrategy.sol";
 contract TestStrategy is BaseStrategy {
     bool public doReentrancy;
 
+    // Some token that needs to be protected for some reason
+    // Initialize this to some fake address, because we're just using it
+    // to test `BaseStrategy.protectedTokens()`
+    address public constant protectedToken = address(0xbad);
+
     constructor(address _vault) public BaseStrategy(_vault) {}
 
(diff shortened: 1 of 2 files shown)


contracts/test/TestStrategy.sol
```diff
@@ -12,6 +12,11 @@ import {BaseStrategy, StrategyParams, VaultAPI} from "../BaseStrategy.sol";
 contract TestStrategy is BaseStrategy {
     bool public doReentrancy;
 
+    // Some token that needs to be protected for some reason
+    // Initialize this to some fake address, because we're just using it
+    // to test `BaseStrategy.protectedTokens()`
+    address public constant protectedToken = address(0xbad);
```
snorkr4, line 18: I'm not sure if I should be using a made up address literal here, or deploying an actual ERC20 token. 

I thought about the latter, but it requires some extra code on the python test side of things, so I wasn't sure if that made sense to do. Plus, in a real example, I assume this would actually be a literal (like [here](https://etherscan.io/address/0xCa8C5e51e235EF1018B2488e4e78e9205064D736#code)).

Or maybe there's a better way to do this?

fubuloubu: If it works, it works!

fubuloubu (approved):

PR #188, opened 2021-02-07
feat: add etherscan auto contract auto verification

steffenix:


steffenix:
Got:

```
TypeError: Etherscan does not support API verification of source code for vyper contracts. You need to verify the source manually.
```

PR #203, opened 2021-02-16
ci: fix functional CI

steffenix:


fubuloubu (approved):
Nice! Seems to have cut CI time in half!

steffenix:
To speed up even more, we would have to split the functional-vault tests

PR #204, opened 2021-02-16
Make all performance and mgmt fees strategy-specific

TheRyanMiller:
Improve vault composability by making fees strategy-specific as outlined in [this post](https://hackmd.io/@Ap_76vwNTg-vxJxbiaLMMQ/BJG3NYDWO).


fubuloubu:
Might reopen in the future, but for now we are going a different ditection fot what this OR is supposed to enable

fubuloubu:
Retrying this approach in spirit here: #253

PR #210, opened 2021-02-17
ci: use pytest-split

steffenix:
After a lot of trial, it's finally ready.

It will check if the file `.test_durations` exists in order to run all the tests or the tests spitted.

The cache will expire every month and it will have to be rebuilt by the first job that runs that month.


.github/workflows/board.yaml
```diff
@@ -0,0 +1,18 @@
+name: Project Board
```
fubuloubu, line 1: Why is this file udpated?

.github/workflows/test.yaml
```diff
@@ -8,86 +8,22 @@ on:
...
-      run: pip install -r requirements-dev.txt
-
-    - name: Compile Code
-      run: brownie compile --size
-
-    - name: Run Tests
-      run: brownie test tests/functional/registry --gas --coverage --revert-tb
-
-  functional-strategy:
+  functional-splitted:
```
fubuloubu, line 12: Just `functional`

fubuloubu (approved):

.github/workflows/board.yaml
steffenix: File is a duplicate from board.yml I have removed it.

PR #222, opened 2021-02-24
Test multiple decimals

### LINKED ISSUE #220: Parametrize test suite on Tokens with different decimal values
We support a number of different tokens using different decimal amounts (e.g. GUSD has 2, USDC has 6, wBTC has 8, etc.), but we only currently test with 18 decimals. So, we should parametrize our Tokens to try different values just to make sure our calculations are consistent enough not to cause problems.

Also need some test cases that ensure precision loss is bounded below some expected amount.

steffenix:
fixes #220

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -21,9 +21,7 @@ def management(accounts):
     yield accounts[3]
 
 
-@pytest.fixture(
-    params=[("Normal", 18), ("NoReturn", 18), ("Normal", 8), ("NoReturn", 8)]
-)
+@pytest.fixture(params=[("Normal", 18), ("NoReturn", 18), ("Normal", 8), ("Normal", 2)])
 def token(gov, Token, request):
     (behaviour, decimal) = request.param
 
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -21,11 +21,15 @@ def management(accounts):
     yield accounts[3]
 
 
-@pytest.fixture(params=["Normal", "NoReturn"])
+@pytest.fixture(
+    params=[("Normal", 18), ("NoReturn", 18), ("Normal", 8), ("NoReturn", 8)]
+)
 def token(gov, Token, request):
-    token = gov.deploy(Token)
+    (behaviour, decimal) = request.param
+
+    token = gov.deploy(Token, decimal)
     # NOTE: Run our test suite using both compliant and non-compliant ERC20 Token
-    if request.param == "NoReturn":
+    if behaviour == "NoReturn":
         token._initialized = False  # otherwise Brownie throws an `AttributeError`
         setattr(token, "transfer", token.transferWithoutReturn)
         setattr(token, "transferFrom", token.transferFromWithoutReturn)
@@ -40,6 +44,7 @@ def vault(gov, guardian, management, token, rewards, Vault):
     vault.initialize(token, gov, rewards, "", "", guardian)
     vault.setDepositLimit(2 ** 256 - 1, {"from": gov})
     vault.setManagement(management, {"from": gov})
+
     # Make it so vault has some AUM to start
     token.approve(vault, token.balanceOf(gov) // 2, {"from": gov})
     vault.deposit(token.balanceOf(gov) // 2, {"from": gov})
--- a/tests/functional/registry/conftest.py
+++ b/tests/functional/registry/conftest.py
@@ -3,17 +3,17 @@ import pytest
 
 @pytest.fixture
 def create_token(gov, Token):
-    def create_token(decimals):
+    def create_token(decimals=18):
         return Token.deploy(decimals, {"from": gov})
 
     yield create_token
 
 
-@pytest.fixture(params=[18, 8])
-def create_vault(gov, create_token, patch_vault_version, request):
+@pytest.fixture
+def create_vault(gov, create_token, patch_vault_version):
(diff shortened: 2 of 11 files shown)


tests/functional/conftest.py
```diff
@@ -21,11 +21,15 @@ def management(accounts):
     yield accounts[3]
 
 
-@pytest.fixture(params=["Normal", "NoReturn"])
+@pytest.fixture(
+    params=[("Normal", 18), ("NoReturn", 18), ("Normal", 8), ("NoReturn", 8)]
```
fubuloubu, line 25: I don't think we need to check `NoReturn` more than once, the behavior of that and decimals should be orthoganal. Also added a much lower value for decimals (like GUSD)
```suggestion
    params=[("Normal", 18), ("NoReturn", 18), ("Normal", 8), ("Normal", 2)]
```

tests/functional/registry/conftest.py
```diff
@@ -3,17 +3,17 @@
...
-        return Token.deploy({"from": gov})
+    def create_token(decimals):
+        return Token.deploy(decimals, {"from": gov})
 
     yield create_token
 
 
-@pytest.fixture
-def create_vault(gov, create_token, patch_vault_version):
+@pytest.fixture(params=[18, 8])
```
fubuloubu, line 12: This shouldn't need it, the registry is not testing this behavior directly.

tests/functional/registry/conftest.py
```diff
@@ -3,17 +3,17 @@
 
 @pytest.fixture
 def create_token(gov, Token):
-    def create_token():
-        return Token.deploy({"from": gov})
+    def create_token(decimals):
```
fubuloubu, line 6: Do this instead, and you don't need to change so much
```suggestion
    def create_token(decimals=18):
```

fubuloubu (approved):

tests/functional/registry/conftest.py
```diff
@@ -3,14 +3,14 @@
...
-        return Token.deploy({"from": gov})
+    def create_token(decimals=18):
+        return Token.deploy(decimals, {"from": gov})
 
     yield create_token
 
 
 @pytest.fixture
-def create_vault(gov, create_token, patch_vault_version):
+def create_vault(gov, create_token, patch_vault_version, request):
```
fubuloubu, line 13: Undo this
```suggestion
def create_vault(gov, create_token, patch_vault_version):
```

tests/functional/vault/test_withdrawal.py
```diff
@@ -306,7 +306,7 @@ def test_profit_degration(chain, gov, token, vault, strategy, rando):
 
     pricePerShareBefore = vault.pricePerShare()
 
-    chain.sleep(10)
+    chain.sleep(1_000)
```
fubuloubu, line 309: Why'd this change?

steffenix: The value isn't changing enough if you use a token with only two decimal so vault.pricePerShare()  == pricePerShareBefore

fubuloubu: was it not passing? or was it not precise enough to show any difference?

steffenix: It's not passing because of the assert a line after

fubuloubu: Can you explain in greater detail? It sounds like you modified a test to get it to pass. What am I missing about why this change is justifiable?

steffenix:
```
@view
@internal
def _shareValue(shares: uint256) -> uint256:
    lockedFundsRatio: uint256 = (block.timestamp - self.lastReport) * self.lockedProfitDegration
    freeFunds: uint256 = self._totalAssets()

    if(lockedFundsRatio < DEGREDATION_COEFFICIENT):
        freeFunds -= (self.lockedProfit - (lockedFundsRatio * self.lockedProfit / DEGREDATION_COEFFICIENT))
    # NOTE: using 1e3 for extra precision here, when decimals is low
    return ((10 ** 3 * (shares * freeFunds)) / self.totalSupply) / 10 ** 3

```

Constants:
```
DEGREDATION_COEFFICIENT = 10 ** 18
lockedProfitDegration = DEGREDATION_COEFFICIENT * 46 /10 ** 6 = 4.6e+13
```
At the moment of the `    pricePerShareBefore = vault.pricePerShare()` is called at least a second has passed since harvest was called.

```
lockedFundsRatio = 1 * 4.6e+13 = 4.6e+13
freeFunds = 100_000
lockedProfit = 10

(10 - (4.6e+13 * 10 / 10 ** 18)) = 9.99954
Because of the rounding we get:
freeFunds -= 9
```

And then ten seconds after the withdrawal we have:
```
lockedFundsRatio = 10 * 4.6e+13 = 4.6e+14
freeFunds = 100_000
lockedProfit = 10

(10 - (4.6e+14 * 10 / 10 ** 18)) = 9.9954
Because of the rounding we get:
freeFunds -= 9
```

So no more funds are freed between the two calls.

fubuloubu: So to summarize, it's just not waiting long enough to see an effect when `decimals` is a lower number? e.g. There is no losses occurring, just not enough time to see gains

steffenix: There is not enough time to see  difference because of the rounding of the lock degradation.

PR #225, opened 2021-03-01
fix: use proper contract for non-compliant ERC20 testing

fubuloubu:
Previous testing did not actually perform the mechanics that non-compliant tokens like Tether have.


fubuloubu:
fixes made in ca36f0113e32abb78e44f61fa95c0f1258daeffd

PR #226, opened 2021-03-01
test: increase test suite coverage to 100% (#156)

x48-crypto:
This PR brings test suite coverage to 100%. It should be noted that BaseStrategy is under-reporting due to a Brownie error related to modifiers (https://github.com/eth-brownie/brownie/issues/809), however I've ensured that even the methods that report as not fully covered are indeed covered.

I've rebased the PR against @snorkr4's prior work as well (https://github.com/iearn-finance/yearn-vaults/pull/200).

IMAGE 2021-03-01 03:59:07
IMAGE 2021-03-01 03:59:09


brownie-config.yaml
```diff
@@ -21,3 +21,5 @@ reports:
     - contracts/test/TestGuestList.sol
   exclude_contracts:
     - SafeMath
+    - SafeERC20
+    - Address
```
x48-crypto, line 25: Removed SafeERC20 and Address from coverage report as they are not a Yearn contracts (they're from OpenZeppelin)

contracts/Vault.vy
```diff
@@ -444,7 +444,7 @@ def setLockedProfitDegration(degration: uint256):
     @param degration The rate of degration in percent per second scaled to 1e18.
     """
     assert msg.sender == self.governance
-    assert degration >= 0 and degration <= DEGREDATION_COEFFICIENT
+    assert degration <= DEGREDATION_COEFFICIENT
```
x48-crypto, line 447: Due to the way the this is structured it's not possible to obtain full coverage for this line. Since "degration" is a uint256 it can never be less than zero (you will get an outside of range error if you try). The only way to obtain full coverage here is to change the assert statement in Vault.vy.

contracts/Vault.vy
```diff
@@ -1331,13 +1331,16 @@ def addStrategyToQueue(strategy: address):
     assert msg.sender in [self.management, self.governance]
     # Must be a current Strategy
     assert self.strategies[strategy].activation > 0
-    # Check if queue is full
-    assert self.withdrawalQueue[MAXIMUM_STRATEGIES - 1] == ZERO_ADDRESS
     # Can't already be in the queue
+    last_idx: uint256 = 0
```
x48-crypto, line 1335: Updated strategy queue code per @fubuloubu . New code is slightly more gas efficient and is fully testable

contracts/test/TestStrategy.sol
```diff
@@ -25,13 +26,18 @@ contract TestStrategy is BaseStrategyInitializable {
 
     // NOTE: This is a test-only function to simulate losses
     function _takeFunds(uint256 amount) public {
-        want.transfer(msg.sender, amount);
+        want.safeTransfer(msg.sender, amount);
```
x48-crypto, line 29: This change is already merged

contracts/test/Token.sol
```diff
@@ -23,20 +24,73 @@ contract Token is ERC20 {
         require(!_blocked[to], "Token transfer refused. Receiver is on blacklist");
         super._beforeTokenTransfer(from, to, amount);
     }
+}
+
+contract TokenNoReturn {
```
x48-crypto, line 29: This is required to get full coverage on erc20_safe_transfer and erc20_safe_transferFrom

tests/functional/conftest.py
```diff
@@ -1,5 +1,7 @@
 import pytest
 
+from brownie import Token, TokenNoReturn
```
x48-crypto, line 3: These changes are already merged to master

tests/functional/strategy/test_misc.py
```diff
@@ -81,6 +81,39 @@ def test_harvest_tend_trigger(chain, gov, vault, token, TestStrategy):
     assert strategy.estimatedTotalAssets() == 0
 
 
+def test_sweep_authority(
```
x48-crypto, line 84: This test is for the "authority" modifier, although the test is not correctly picked up by brownie due to https://github.com/eth-brownie/brownie/issues/809

tests/functional/vault/test_config.py
```diff
@@ -52,14 +53,25 @@ def test_vault_name_symbol_override(guardian, gov, rewards, token, Vault):
...
     "getter,setter,val,guard_allowed",
     [
         ("name", "setName", "NewName yVault", False),
         ("symbol", "setSymbol", "yvNEW", False),
         ("emergencyShutdown", "setEmergencyShutdown", True, True),
+        ("emergencyShutdown", "setEmergencyShutdown", False, False),
         ("guardian", "setGuardian", None, True),
+        ("guestList", "setGuestList", None, False),
         ("rewards", "setRewards", None, False),
+        ("lockedProfitDegration", "setLockedProfitDegration", 1000, False),
```
x48-crypto, line 74: I believe this was added by @snorkr4

contracts/Vault.vy
fubuloubu: Add a comment above this

contracts/Vault.vy
fubuloubu: Should result in 1 less `SLOAD` (note currently iteration over `self.withdrawalQueue` just loads the whole thing anyways)

contracts/test/TestStrategy.sol
```diff
@@ -85,8 +91,7 @@ contract TestStrategy is BaseStrategyInitializable {
         } else {
             // NOTE: Just in case something was stolen from this contract
             if (totalDebt > totalAssets) {
-                _loss = totalDebt.sub(totalAssets);
-                if (_loss > _amountNeeded) _loss = _amountNeeded;
+                _loss = _amountNeeded;
```
fubuloubu, line 94: I don't think this is correct

tests/functional/vault/test_config.py
```diff
@@ -177,3 +189,11 @@ def test_vault_setGovernance(gov, vault, rando):
...
     with brownie.reverts():
         vault.acceptGovernance({"from": gov})
+
+
+def test_vault_setLockedProfitDegration_range(gov, vault):
+    # value must be between 0 and DEGREDATION_COEFFICIENT (inclusive)
+    vault.setLockedProfitDegration(0, {"from": gov})
+    vault.setLockedProfitDegration(DEGREDATION_COEFFICIENT, {"from": gov})
+    with brownie.reverts():
+        vault.setLockedProfitDegration(DEGREDATION_COEFFICIENT + 1000, {"from": gov})
```
fubuloubu, line 199: Show off-by-one behavior correct
```suggestion
        vault.setLockedProfitDegration(DEGREDATION_COEFFICIENT + 1, {"from": gov})
```

tests/functional/vault/test_misc.py
```diff
@@ -2,12 +2,97 @@
...
 def other_token(gov, Token):
     yield gov.deploy(Token, 18)
 
 
+@pytest.fixture
+def token_false_return(gov, TokenFalseReturn):
+    yield gov.deploy(TokenFalseReturn, 18)
+
+
+def vault(gov, management, token, Vault):
```
fubuloubu, line 19: Missing `fixture`?

tests/functional/vault/test_strategies.py
```diff
@@ -30,29 +28,59 @@ def other_strategy(gov, vault, TestStrategy):
...
     # Only governance can add a strategy
     with brownie.reverts():
         vault.addStrategy(strategy, 100, 10, 20, 1000, {"from": rando})
 
+    # Can't add a strategy during emergency shutdown
+    vault.setEmergencyShutdown(True, {"from": gov})
+    with brownie.reverts():
+        vault.addStrategy(strategy, 100, 10, 20, 1000, {"from": gov})
+    chain.undo()
+    chain.undo()
```
fubuloubu, line 82: Why undo twice? this puts it before the test?

snorkr4: I can't find any documentation on this, but as far as I can tell from experimentation - at line 81, the `addStrategy` call from line 80 is still in the transaction history. That's why I added the second undo.

tests/functional/vault/test_strategies.py
```diff
@@ -314,8 +354,42 @@ def test_ordering(gov, vault, TestStrategy, rando):
...
+
+    # Not just anyone can add a strategy to the queue
+    with brownie.reverts():
+      vault.addStrategyToQueue(strategy, {"from": rando})
+
+    # # Governance can add a strategy to the queue
+    vault.addStrategyToQueue(strategy, {"from": gov})
+    vault.removeStrategyFromQueue(strategy, {"from": gov})
+      
+    # Managemenet can add a strategy to the queue
```
snorkr4, line 375: nit: typo
```suggestion
    # Management can add a strategy to the queue
```

tests/functional/vault/test_strategies.py
```diff
@@ -314,8 +354,42 @@ def test_ordering(gov, vault, TestStrategy, rando):
     with brownie.reverts():
         vault.addStrategyToQueue(removed_strategy, {"from": gov})
 
+def test_add_strategy_to_queue(gov, management, vault, TestStrategy, strategy, other_strategy, rando):
```
snorkr4, line 357: nit: to be stylistically consistent with `test_addStrategy`, I guess this could be called `test_addStrategyToQueue`?

tests/functional/vault/test_strategies.py
fubuloubu: I don't think that is the case, but if you were having issues with it we should at least add a comment here as to what is being rolled back with each call.

snorkr4 (commented):
Thanks for tagging me and including my changes. Was helpful to me to read through this.

contracts/test/TestStrategy.sol
x48-crypto: In my testing I've not been able to get `_loss > _amountNeeded` to return false. I believe this line is over-defined/always true, so I set `_loss = _amountNeeded` as this is what happens when `_loss > _amountNeeded` is true.

tests/functional/vault/test_strategies.py
snorkr4, line 82:
```suggestion
    # Even if the previous call to vault.addStrategy reverts, it does not 
    # rollback, so we must undo twice to undo the emergency shutdown.
    chain.undo() # vault.addStrategy
    chain.undo() # vault.setEmergencyShutdown
```

x48-crypto: @snorkr4 what would you suggest for a comment here?

x48-crypto: @snorkr4 nvm, saw your suggestion and merged. thanks!

tests/functional/vault/test_misc.py
x48-crypto: @snorkr4 do you know why this is by any chance? I believe this came from your PR originally. I noticed if I add the fixture the test fails

contracts/test/TestStrategy.sol
x48-crypto: I've reverted this commit as I was able to add a new unit test to obtain full coverage without modifying this code

image
image

fubuloubu:
Fix commitlint and run `black` on changes

fubuloubu (approved):

tests/functional/vault/test_misc.py
x48-crypto: I've made `vault` a fixture and fixed the broken unit tests

x48-crypto:
All issues have been resolved and the PR has been rebased against master.

tests/functional/vault/test_misc.py
```diff
@@ -49,18 +50,19 @@ def test_credit_available_minDebtPerHarvest_larger_than_available(
...
-    vault.setDepositLimit(2 ** 256 - 1, {"from": gov})
+    vault.setDepositLimit(MAX_UINT256, {"from": gov})
     strategy = gov.deploy(TestStrategy, vault)
+    minDebtPerHarvest = MAX_UINT256 - 1
     vault.addStrategy(
         strategy,
         10000,  # 100% of Vault AUM
-        2 ** 256 - 2,  # minDebtPerHarvest
-        2 ** 256 - 1,  # maxDebtPerHarvest
+        minDebtPerHarvest,  # minDebtPerHarvest
```
fubuloubu, line 59: why is this set so high? (can it be 0)?

x48-crypto: In order to test the line `available < strategy_minDebtPerHarvest` `minDebtToHarvest` must be greater than `available` (it cannot be zero).

Previously I had set `minDebtToHarvest` to `MAX_UINT256 - 1` to ensure this line gets tested correctly, but I've updated the PR and now I start with `minDebtToHarvest` set to zero and update it to `available + 1`.

It should be noted that this test is specifically to test that `creditAvailable` returns zero after a liquidation when `minDebtToHarvest` is greater than `available`. We do have another unit test that covers liquidation where `minDebtPerHarvest` is zero (or less than `available`).

This is the primary assert we are testing for:
```
    minDebtPerHarvestExceedsAvailable = strategy_minDebtPerHarvest > available
    assert minDebtPerHarvestExceedsAvailable == True

    creditAvalable = vault.creditAvailable(strategy)
    assert creditAvalable == 0
```

This is the code we are adding coverage for:
image

fubuloubu: Sounds good!

PR #228, opened 2021-03-01
fix gov & guardian role in deployment

orbxball:


scripts/deploy.py
```diff
@@ -62,12 +62,12 @@ def main():
         gov_default = dev.address
     else:
         gov_default = "ychad.eth"
-    gov = get_address("Yearn Governance", default=gov_default)
+    gov = get_address("Yearn Governance", default="0x16388463d60FFE0661Cf7F1f31a7D658aC790ff7")
```
fubuloubu, line 65: What is this address?

orbxball: strategist multisig.
these two changes are to match the latest version of `OPERATION.md`

scripts/deploy.py
```diff
@@ -62,12 +62,12 @@ def main():
...
 
     rewards = get_address(
         "Rewards contract", default="0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde"
```
fubuloubu, line 68:
```suggestion
        "Rewards contract", default="treasury.ychad.eth"
```

scripts/deploy.py
fubuloubu: Please add an ENS entry for this msig and reference that instead

orbxball: I don think we have ENS on strategist msig. But it's fine. We agreed to set `gov = dev.ychad.eth` on experimental vault.

fubuloubu: that's a big change, to me... I was expecting strategists would gov experiments because the whole idea is that experiments drive production strategies and vaults. devs are in charge of the production protocol, strategists in charge of experiments.

fameal: This was a long discussion and we finally agreed on ydev governs production vaults. Other stuff is up to strategists. We can keep this as it was.

scripts/deploy.py
```diff
@@ -62,12 +62,12 @@ def main():
         gov_default = dev.address
     else:
         gov_default = "ychad.eth"
-    gov = get_address("Yearn Governance", default=gov_default)
+    gov = get_address("Yearn Governance", default="dev.ychad.eth")
```
fubuloubu, line 65: Revert this change back to `gov_default`

fubuloubu (approved):

scripts/deploy.py
```diff
@@ -62,12 +62,12 @@ def main():
...
 
     rewards = get_address(
-        "Rewards contract", default="0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde"
+        "Rewards contract", default="treasury.ychad.eth"
```
fubuloubu, line 68: Run `black .` on this

fubuloubu:
Please rebase on `master`

scripts/deploy.py
```diff
@@ -59,15 +59,15 @@ def main():
     token = Token.at(get_address("ERC20 Token"))
 
     if use_proxy:
-        gov_default = dev.address
+        gov_default = (
+            "0x16388463d60FFE0661Cf7F1f31a7D658aC790ff7"  # strategist msig, no ENS
```
fubuloubu, line 63: can the strategist msig be added to ENS? e.g. `brains.ychad.eth`

orbxball: It's on Klim's long list already.

PR #234, opened 2021-03-08
feat: deploy proxy with prev release

fubuloubu:


banteg (approved):

contracts/Registry.vy
```diff
@@ -176,9 +176,9 @@ def _newProxyVault(
     guardian: address,
     name: String[64],
     symbol: String[32],
+    releaseTarget: uint256,
 ) -> address:
-    # NOTE: Underflow if no releases created yet (this is okay)
-    vault: address = create_forwarder_to(self.releases[self.nextRelease - 1])  # dev: no releases
+    vault: address = create_forwarder_to(self.releases[releaseTarget])
```
dougstorm, line 181: what happens if releaseTarget doesnt exist?

contracts/Registry.vy
```diff
@@ -269,14 +277,14 @@ def endorseVault(vault: address):
...
     assert Vault(vault).governance() == msg.sender  # dev: not governed
 
-    # NOTE: Underflow if no releases created yet (this is okay)
-    api_version: String[28] = (
-        Vault(self.releases[self.nextRelease - 1]).apiVersion()  # dev: no releases
-    )
+    # NOTE: Underflow if no releases created yet, or targeting prior to release history
+    releaseTarget: uint256 = self.nextRelease - 1 - releaseDelta  # dev: no releases
+    api_version: String[28] = Vault(self.releases[releaseTarget]).apiVersion()
     assert Vault(vault).apiVersion() == api_version  # dev: not latest release
```
dougstorm, line 288: maybe update comment that apiVersion does not match target release?

contracts/Registry.vy
fubuloubu: It should fail when it does checks further down the line, but easy to check upfront

dougstorm: nvm, i noticed this is internal so the delta protects against using release targets that we don't have

dougstorm (approved):
LGTM, only one minor comment

contracts/Registry.vy
fubuloubu: fixed

contracts/Registry.vy
fubuloubu: fixed

PR #235, opened 2021-03-09
fix: allow for some small loss on migration, but disallow underflows

fubuloubu:
Failure found in https://github.com/yearn/yearn-vaults/runs/2060926601


dougstorm (approved):
LGTM!

PR #236, opened 2021-03-09
feat(registry): add ability to track endorsed vault tokens

x48-crypto:
Added the ability to track endorsed vault tokens in the registry.

- Only support adding tokens if they've not yet been added
- Only keep track of tokens for endorsed vaults (deployments)

Screen Shot 2021-03-10 at 7 51 00 PM

### DIFF
--- a/contracts/Registry.vy
+++ b/contracts/Registry.vy
@@ -138,16 +138,6 @@ def _registerRelease(vault: address):
 
 
 @internal
-def _registerToken(token: address):
-    tokenIsRegistered: bool = self.isRegistered[token]
-    if tokenIsRegistered:
-      return
-    self.tokens[self.numTokens] = token
-    self.isRegistered[token] = True
-    self.numTokens = self.numTokens + 1
-
-
-@internal
 def _registerDeployment(token: address, vault: address):
     # Check if there is an existing deployment for this token at the particular api version
     # NOTE: This doesn't check for strict semver-style linearly increasing release versions
@@ -164,7 +154,10 @@ def _registerDeployment(token: address, vault: address):
     self.nextDeployment[token] = deployment_id + 1
 
     # Register tokens for endorsed vaults
-    self._registerToken(token)
+    if not self.isRegistered[token]:
+        self.isRegistered[token] = True
+        self.tokens[self.numTokens] = token
+        self.numTokens += 1
     
     # Log the deployment for external listeners (e.g. Graph)
     log NewVault(token, deployment_id, vault, Vault(vault).apiVersion())
--- a/contracts/Registry.vy
+++ b/contracts/Registry.vy
@@ -23,10 +23,14 @@ releases: public(HashMap[uint256, address])
 nextDeployment: public(HashMap[address, uint256])
 vaults: public(HashMap[address, HashMap[uint256, address]])
 
-# Maintain a list of endorsed vault tokens
-tokensList: public(address[65536])
-tokensMap: public(HashMap[address, bool])
-tokensCount: public(uint256)
+# Index of token added => token address
+tokens: public(HashMap[uint256, address])
+
+# len(tokens)
+numTokens: public(uint256)
+
+# Inclusion check for token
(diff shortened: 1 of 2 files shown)


contracts/Registry.vy
```diff
@@ -128,6 +133,16 @@ def _registerRelease(vault: address):
     log NewRelease(release_id, vault, Vault(vault).apiVersion())
 
 
+@internal
+def _registerToken(token: address):
```
x48-crypto, line 137: I considered adding a check here to make sure the token that's being added maps to a deployed vault, but since this check is already done in _registerDeployment it seemed like this check would be redundant and perhaps a waste of gas.

As long as _registerToken is only called from _registerDeployment (this is currently the case) this should not be a problem.

Alternatively, we can remove the internal method and move lines 138-143 to the bottom of _registerDeployment.

We could also keep this commit how it is as I believe the current version is more readable than the proposed alternative.

contracts/Registry.vy
```diff
@@ -23,6 +23,11 @@ releases: public(HashMap[uint256, address])
 nextDeployment: public(HashMap[address, uint256])
 vaults: public(HashMap[address, HashMap[uint256, address]])
 
+# Maintain a list of endorsed vault tokens
+tokensList: public(address[65536])
+tokensMap: public(HashMap[address, bool])
+tokensCount: public(uint256)
```
fubuloubu, line 29: refactor to:
```python
# index of token added => token address
tokens: public(HashMap[uint256, address])
# len(tokens)
numTokens: public(uint256)
# inclusion check for token
isRegistered: public(HashMap[address, bool])
```

fubuloubu (approved):
Just want to refactor it to something more scalable. `tokensList` will cost a lot in gas to update.

contracts/Registry.vy
x48-crypto: I added the recommended fix and updated the unit tests as well

contracts/Registry.vy
x48-crypto: Moved _registerToken logic into _registerDeployment

contracts/Registry.vy
```diff
@@ -128,6 +137,16 @@ def _registerRelease(vault: address):
...
 
 
+@internal
+def _registerToken(token: address):
+    tokenIsRegistered: bool = self.isRegistered[token]
+    if tokenIsRegistered:
+      return
+    self.tokens[self.numTokens] = token
+    self.isRegistered[token] = True
+    self.numTokens = self.numTokens + 1
```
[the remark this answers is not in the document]
x48-crypto: Implemented this suggestion

contracts/Registry.vy
```diff
@@ -144,6 +163,9 @@ def _registerDeployment(token: address, vault: address):
     self.vaults[token][deployment_id] = vault
     self.nextDeployment[token] = deployment_id + 1
 
+    # Register tokens for endorsed vaults
+    self._registerToken(token)
```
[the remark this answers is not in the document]
x48-crypto: Moved _registerToken logic into _registerDeployment

contracts/Registry.vy
```diff
@@ -23,6 +23,15 @@ releases: public(HashMap[uint256, address])
...
 vaults: public(HashMap[address, HashMap[uint256, address]])
 
+# Index of token added => token address
+tokens: public(HashMap[uint256, address])
+
+# len(tokens)
+numTokens: public(uint256)
+
+# Inclusion check for token
+isRegistered: public(HashMap[address, bool])
```
fubuloubu, line 33: I like grouping related things together here
```suggestion
# Index of token added => token address
tokens: public(HashMap[uint256, address])
# len(tokens)
numTokens: public(uint256)
# Inclusion check for token
isRegistered: public(HashMap[address, bool])
```

fubuloubu (approved):

fubuloubu:
#237 fixes the failure raised in CI

PR #237, opened 2021-03-10
fix: cheapen wrapper withdrawals

fubuloubu:

PR #238, opened 2021-03-10
refactor: standardize terminology for vaults in registry

fubuloubu:
This would have driven me crazy if we didn't standardize this before redeploying


banteg (approved):

fubuloubu:
#237 fixes the failure raised in CI

PR #243, opened 2021-03-14
feat: remove protected tokens

### LINKED ISSUE #241: Remove protectedTokens
protectedTokens serves no purpose and is in fact dangerous. It is there to stop governance from 'sweeping' out want and other protected tokens. But why do we want this? It adds no extra safety as governance has that ability with migrateStrategy. But if migrateStrategy is broken then we could end up with tokens locked forever.

steffenix:
closes #241

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -770,25 +770,6 @@ abstract contract BaseStrategy {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


fubuloubu:
Closing per https://github.com/yearn/yearn-vaults/issues/241#issuecomment-802858038

PR #245, opened 2021-03-15
fix: don't check governance on uninitialized vault

fubuloubu:


contracts/Registry.vy
```diff
@@ -115,8 +115,22 @@ def latestVault(token: address) -> address:
...
+        deployed from the registry throw other methods (to save gas).
+    @dev
+        Throws if caller isn't `self.governance`.
+        Throws if `vault`'s governance isn't `self.governance`.
+        Throws if the api version is the same as the previous release.
+        Emits a `NewVault` event.
+    @param vault The vault that will be used as the template contract for the next release.
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+
```
dougstorm, line 133: sorry, i missed a bit of context, what was the cause for removing the governance check in the release vault?

dougstorm: since we are using a proxy i can't think of a way to take advantage of a non-initialised vault as backing implementation, but it still feels off that we leave empty for any other user to initalize the implementation.

We can force to initialize to ychad.eth and stub data the implementation to brick the release just in case

fubuloubu: `vault` might not be initialized, and since we removed the need to endorse, there's no reason to check for an initialized parameter any more.

`apiVersion` is present in an uninitialized Vault release

dougstorm (approved):
LGTM!

Issue #247, opened 2021-03-16
Withdrawal of 0 shares is allowed and not reverted


jmonteer:
Two scenarios: 
A user with no shares tries to withdraw full balance (= 0)
A user tries to withdraw 0 shares explicitly

Withdrawal should revert when the users tries to (implicitly or explicitly) withdraw 0 shares.

This is to be consistent with the approach used in deposit function (https://github.com/yearn/yearn-vaults/blob/e390c2a6b2ba6e2ecc8f3a72a1ea4adfabd17544/contracts/Vault.vy#L856) 

It is not critical as it does not lock anybody in the vault or implies any relevant error.

jmonteer:
closed via #248

Issue #246, opened 2021-03-16
Vault withdrawal reverts with "division by zero" when totalSupply == 0


jmonteer:
When vault is empty and totalSupply == 0, withdrawal reverts with division by zero.

This does not affect expected behavior in the sense that it does not lock anybody in the vault (the withdrawal should revert anyway as the user will not own any shares) but it should not return "div by zero".

PR #248, opened 2021-03-16
fix: issues with withdrawal from empty vaults and 0 shares withdrawals

jmonteer:
Refer to issues #247 and #246 

Added a shortcut in _shareValue to return 1:1 sharePrice if totalSupply == 0 to avoid division by zero

Checks that the user is withdrawing > 0 

Bonus: Checks that maxLoss is <= MAX_BPS


fubuloubu (approved):
Add tests for new `assert` clauses

tests/functional/vault/test_withdrawal.py
```diff
@@ -315,3 +315,37 @@ def test_profit_degration(chain, gov, token, vault, strategy, rando):
...
+    assert token.balanceOf(gov) == 0
+
+    # Deposit something in strategies
+    chain.sleep(1)  # Needs to be a second ahead, at least
+    [s.harvest({"from": gov}) for s in strategies]
+    assert token.balanceOf(vault) < vault.totalAssets()  # Some debt is in strategies
+
+    # Trying to withdraw 0 shares. It should revert
+    with brownie.reverts():
+        vault.withdraw(0, {"from": rando})
```
fubuloubu, line 348: This line could probably just be in another test

contracts/Vault.vy
```diff
@@ -972,13 +976,19 @@ def withdraw(
     """
     shares: uint256 = maxShares  # May reduce this number below
 
+    # Max Loss is <=100%, revert otherwise
+    assert maxLoss <= MAX_BPS
```
fubuloubu, line 980: missing a test case showing this

contracts/Vault.vy
```diff
@@ -1058,10 +1068,7 @@ def pricePerShare() -> uint256:
     @dev See dev note on `withdraw`.
     @return The value of a single share.
     """
-    if self.totalSupply == 0:
-        return 10 ** self.decimals  # price of 1:1
-    else:
-        return self._shareValue(10 ** self.decimals)
+    return self._shareValue(10 ** self.decimals)
```
poolpitako, line 1071: revert

poolpitako (approved):

tests/functional/vault/test_withdrawal.py
jmonteer: merging with test_progressive_withdrawal as suggested

jmonteer:
it should be good to go @fubuloubu

contracts/BaseWrapper.sol
```diff
@@ -179,7 +179,8 @@ abstract contract BaseWrapper {
 
                     // Limit amount to withdraw to the maximum made available to this contract
                     uint256 shares = Math.min(estimatedShares, availableShares);
-                    withdrawn = withdrawn.add(vaults[id].withdraw(shares));
+                    // If shares to be withdrawn is 0, do not try to withdraw (as it will revert)
+                    withdrawn = shares == 0 ? withdrawn : withdrawn.add(vaults[id].withdraw(shares));
```
fubuloubu, line 183: `availableShares` is > 0 (see above guard clause), and `amount > withdrawn` should also be true (see below `break` clause)

jmonteer: I thought the same. However for some reason tests were not passing before adding that switch...

contracts/Vault.vy
fubuloubu: they changed `_shareValue` above too

tests/functional/vault/test_withdrawal.py
```diff
@@ -101,6 +101,11 @@ def test_forced_withdrawal(token, gov, vault, TestStrategy, rando, chain):
 
     # Scenario 1: we panic, and try to get out as quickly as possible (total loss)
     assert token.balanceOf(rando) == 0
+
+    # User first try to withdraw with +100% losses, which are not allowed
```
fubuloubu, line 105:
```suggestion
    # User first try to withdraw with more than 100% losses, which is nonsensical
```

contracts/Vault.vy
```diff
@@ -1011,7 +1021,7 @@ def withdraw(
                 continue  # Nothing to withdraw from this Strategy, try the next one
 
             # Force withdraw amount from each Strategy in the order set by governance
-            loss: uint256 = Strategy(strategy).withdraw(amountNeeded)
+            loss: uint256 = Strategy(strategy).withdraw(amountNeeded)
```
fubuloubu, line 1024: undo this change? adds extra white space

tests/integration/test_wrapper.py
```diff
@@ -82,7 +82,6 @@ def rule_withdraw(self, ratio="st_ratio"):
         if amount > 0:
             print(f"  Wrapper.withdraw({amount})")
             self.wrapper.withdraw(amount, {"from": self.user}).return_value
-
```
fubuloubu, line 85: undo this change, makes it easier to read

contracts/BaseWrapper.sol
fubuloubu: Can you point to this issue in a CI failure?

fubuloubu, line 183: Prefer something more like this
```suggestion
                    // If shares to be withdrawn is 0, do not try to withdraw (as it will revert)
                    if (shares > 0) withdrawn = withdrawn.add(vaults[id].withdraw(shares));
```

jmonteer: got it working without this fix.

resolving

jmonteer:
@fubuloubu now it should be ready to go

fubuloubu (approved):

PR #253, opened 2021-03-22
refactor: delegated fees

fubuloubu:
This PR changes how fees are assessed such that `BaseStrategy.delegatedAssets` is pro-rated into the fee assessment. If a strategy delegates all of it's assets, there should be no fees earned by that strategy to either governance or the strategist themselves, and that portion of the assets loaned out shall earn no management fees either.

If only a portion is delegated (e.g. debt-based strategies where some of the funds are used to borrow against), that portion should be calculated such that the exact funds used to borrow are marked as delegated (e.g. Maker strategy w/ 200% debt ratio, then half the funds should be marked "delegated"). This is important to ensure than the degree of work is properly accounted for in how a strategy operates.

Strategists will need to be aware of properly reporting `delegatedAssets` and ensure it is strictly less than or equal to the total debt, or else the calculations will revert or be processed improperly.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1648,7 +1648,11 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     # Update cached value of delegated assets
     #   (used to properly account for mgmt fee in `_assessFees`)
     self.delegatedAssets -= self._strategy_delegatedAssets[msg.sender]
-    delegatedAssets: uint256 = Strategy(msg.sender).delegatedAssets()
+    # NOTE: Use `min(totalDebt, delegatedAssets)` as a guard against improper computation
+    delegatedAssets: uint256 = min(
+        self.strategies[msg.sender].totalDebt,
+        Strategy(msg.sender).delegatedAssets(),
+    )
     self.delegatedAssets += delegatedAssets
     self._strategy_delegatedAssets[msg.sender] = delegatedAssets
 
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -52,6 +52,7 @@ interface Strategy:
     def want() -> address: view
     def vault() -> address: view
     def isActive() -> bool: view
+    def delegatedAssets() -> uint256: view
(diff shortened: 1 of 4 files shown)


contracts/Vault.vy
```diff
@@ -1637,6 +1645,13 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
         self.erc20_safe_transferFrom(self.token.address, msg.sender, self, totalAvail - credit)
     # else, don't do anything because it is balanced
 
+    # Update cached value of delegated assets
+    #   (used to properly account for mgmt fee in `_assessFees`)
+    self.delegatedAssets -= self._strategy_delegatedAssets[msg.sender]
+    delegatedAssets: uint256 = Strategy(msg.sender).delegatedAssets()
```
dougstorm, line 1651: I like that this approach is backward compatible with old base strategies and we can reuse almost all of our previous code

dougstorm:
Thinking if the comment here also needs to document this?

https://github.com/yearn/yearn-vaults/blob/ba980d10c83c5676f2e89ff557b9a753b245b28b/contracts/BaseStrategy.sol#L217

contracts/Vault.vy
```diff
@@ -1637,6 +1645,13 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
...
+    self.delegatedAssets += delegatedAssets
```
dougstorm, line 1652: thinking if we should Min here between `delegatedAssets` and `totalAssets` from strategy, in case of a faulty strategy calc vault can still keep working with less fees

contracts/Vault.vy
fubuloubu: Assuming the strategists filled this value out correctly, which hopefully is a reasonable assumption

contracts/Vault.vy
fubuloubu: Agreed, let's do this

dougstorm (approved):
LGTM on latest changes, all comments addressed

PR #254, opened 2021-03-25
refactor: use `ethToWant` to convert gas costs to want

fubuloubu:


contracts/BaseStrategy.sol
```diff
@@ -466,6 +466,22 @@ abstract contract BaseStrategy {
...
+     *  Care must be taken when working with decimals to assure that the conversion
+     *  is compatible. As an example:
+     *
+     *      given 1e17 wei (0.1 ETH) as input, and want is USDC (6 decimals),
+     *      with USDC/ETH = 1800, this should give back 1800000000 (180 USDC)
+     *
+     * @param _amtInWei The amount (in wei/1e-18 ETH) to convert to `want`
+     * @return The amount in `want` of `_amtInEth` converted to `want`
+     **/
+    function ethToWant(uint256 _amtInWei) public virtual view returns (uint256);
```
poolpitako, line 483: instead of abstract why don't we make the default behavior be 1:1

poolpitako (approved):

contracts/BaseStrategy.sol
fubuloubu: No we want to force people to implement something cc @Grandthrax

poolpitako: ok, fair!

PR #255, opened 2021-03-25
refactor: use `_reportLoss` function on losses to reduce impact

fubuloubu:


poolpitako (approved):

fubuloubu:
cc @jmonteer can you take a look at this fix?

contracts/Vault.vy
```diff
@@ -1029,12 +1046,15 @@ def withdraw(
...
 
             # Reduce the Strategy's debt by the amount withdrawn ("realized returns")
             # NOTE: This doesn't add to returns as it's not earned by "normal means"
-            self.strategies[strategy].totalDebt -= withdrawn + loss
-            self.totalDebt -= withdrawn + loss
+            self.strategies[strategy].totalDebt -= withdrawn
+            totalWithdrawn += withdrawn
+
+        # NOTE: Do this once at the end of the withdrawal queue to reduce `SSTORE` ops
+        self.totalDebt -= totalWithdrawn
```
jmonteer, line 1057: May changing `totalDebt` here cause problems when recalculating the ratio at the moment of reducing our trust with the strategy in line 944 (`self._totalAssets()` uses `self._totalDebt`) if we withdraw from several strategies and some report losses?

this is, if we withdraw from several strategies and some (other than the first) report losses, they will be seeing a higher _totalAssets than they should, resulting in a wrong `ratio_change` calculation

am I missing something?

fubuloubu: Great point! I will remove the gas optimization.

dougstorm (approved):
LGTM!

PR #257, opened 2021-03-29
fix: prevent mgmt fee from creating negative APY situation

TheRyanMiller:
WBTC vault currently shows negative APY because mgmt fee is taking a 2% cut on unprofitable harvests. This fix will prevent mgmt fee from exceeding gains on report. An edge case will still exist where a non-zero profit is reported, but performance fees cause APY to be negative.


fubuloubu:
Similar fix in #263 

Thanks for inspiring us!

PR #258, opened 2021-03-29
fix: delegated assets bug

fubuloubu:


dougstorm (approved):
LGTM, added a simple test to assert accounting of delegated assets is correct after withdrawals

PR #261, opened 2021-04-02
Fix: deposit everything to AffiliateToken

AlexanderMazaletskiy:
First bug:

See:
1.https://github.com/yearn/brownie-wrapper-mix/blob/master/contracts/AffiliateToken.sol#L101

Here share is calculated before the deposit.

2. If we consider the situation in which MAX UINT256 passed as input
https://github.com/yearn/brownie-wrapper-mix/blob/master/contracts/AffiliateToken.sol#L98

3. When we pass the MAX_UINT value, the vulnerability appears in this code.
See:
1.https://github.com/yearn/brownie-wrapper-mix/blob/master/contracts/AffiliateToken.sol#L101

Here share is calculated before the deposit.

2. If we consider the situation in which MAX UINT256 passed as input
https://github.com/yearn/brownie-wrapper-mix/blob/master/contracts/AffiliateToken.sol#L98

3. When we pass the MAX_UINT value, the vulnerability appears in this code.

Since share will be calculated from the MAX_UINT amount
https://github.com/yearn/brownie-wrapper-mix/blob/master/contracts/AffiliateToken.sol#L91, not the actual amount that was sent to the deposit.

Which ultimately leads to the fact that https://github.com/yearn/brownie-wrapper-mix/blob/master/contracts/AffiliateToken.sol#L104

the wrong number of tokens will be issued.

Another bug:

See:

1. Here, pass MAX_UINT to trigger DEPOSIT_EVERTING, in the _deposit function
https://github.com/yearn/brownie-wrapper-mix/blob/master/contracts/AffiliateToken.sol#L98

2. Next,  try to call safeTransferFrom
https://github.com/yearn/yearn-vaults/blob/master/contracts/BaseWrapper.sol#L113

But at the same time we are trying to write off MAX_UINT, of course, which are not on the user's balance.

As a result, it gets that the transaction will fail.

### DIFF
--- a/contracts/test/AffiliateToken.sol
+++ b/contracts/test/AffiliateToken.sol
@@ -64,11 +64,12 @@ contract AffiliateToken is ERC20, BaseWrapper {
     }
 
     function pricePerShare() external view returns (uint256) {
-        return 10**uint256(decimals());
+        return totalVaultBalance(address(this)).mul(10**uint256(decimals())).div(totalSupply());
     }
 
     function _sharesForValue(uint256 amount) internal view returns (uint256) {
-        uint256 totalWrapperAssets = totalVaultBalance(address(this));
+        // total wrapper assets before deposit (assumes deposit already occured)
+        uint256 totalWrapperAssets = totalVaultBalance(address(this)).sub(amount);
 
         if (totalWrapperAssets > 0) {
             return totalSupply().mul(amount).div(totalWrapperAssets);
@@ -82,8 +83,8 @@ contract AffiliateToken is ERC20, BaseWrapper {
     }
 
     function deposit(uint256 amount) public returns (uint256 deposited) {
-        uint256 shares = _sharesForValue(amount); // NOTE: Must be calculated before deposit is handled
         deposited = _deposit(msg.sender, address(this), amount, true); // `true` = pull from `msg.sender`
+        uint256 shares = _sharesForValue(deposited);  // NOTE: Must be calculated after deposit is handled
         _mint(msg.sender, shares);
     }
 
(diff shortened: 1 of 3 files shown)


contracts/test/AffiliateToken.sol
```diff
@@ -64,11 +64,11 @@ contract AffiliateToken is ERC20, BaseWrapper {
     }
 
     function pricePerShare() external view returns (uint256) {
-        return 10**uint256(decimals());
+        return totalVaultBalance(address(this)).mul(10**uint256(decimals())).div(totalSupply());
     }
 
     function _sharesForValue(uint256 amount) internal view returns (uint256) {
-        uint256 totalWrapperAssets = totalVaultBalance(address(this));
+        uint256 totalWrapperAssets = totalVaultBalance(address(this)) - amount; //total wrapper assets before deposit
```
fubuloubu, line 71:
```suggestion
        // total wrapper assets before deposit (assumes deposit already occured)
        uint256 totalWrapperAssets = totalVaultBalance(address(this)).sub(amount);
```

contracts/test/AffiliateToken.sol
```diff
@@ -82,8 +82,8 @@ contract AffiliateToken is ERC20, BaseWrapper {
     }
 
     function deposit(uint256 amount) public returns (uint256 deposited) {
-        uint256 shares = _sharesForValue(amount); // NOTE: Must be calculated before deposit is handled
         deposited = _deposit(msg.sender, address(this), amount, true); // `true` = pull from `msg.sender`
+        uint256 shares = _sharesForValue(deposited);
```
fubuloubu, line 86:
```suggestion
        uint256 shares = _sharesForValue(deposited);  // NOTE: Must be calculated after deposit is handled
```

fubuloubu (approved):
Fix first commit message

AlexanderMazaletskiy:
@fubuloubu I updated PR

PR #263, opened 2021-04-04
feat: locked profit ratio

Grandthrax:
Adds a variable to lock a ratio of profit reported by vault. The purpose is to set a number that stops there being a share price dip after harvest. Currently there is a dip in share price because fees are taken but all profit is locked.


contracts/Vault.vy
```diff
@@ -1674,7 +1689,7 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     # Update reporting time
     self.strategies[msg.sender].lastReport = block.timestamp
     self.lastReport = block.timestamp
-    self.lockedProfit = gain # profit is locked and gradually released per block
+    self.lockedProfit = gain * self.lockedProfitRatio / MAX_BPS  # profit is locked and gradually released per block
```
fubuloubu, line 1692: return `total_fee` from `_assessFees` and modify this to:

```suggestion
    self.lockedProfit = gain - total_fee  # profit is locked and gradually released per block
```

NOTE: `total_fee` must be adjusted to not be greater than `gain`

contracts/Vault.vy
```diff
@@ -1540,20 +1522,21 @@ def _reportLoss(strategy: address, loss: uint256):
     self.debtRatio -= ratio_change
 
 @internal
-def _assessFees(strategy: address, gain: uint256):
+def _assessFees(strategy: address, gain: uint256) -> uint256:
     # Issue new shares to cover fees
     # NOTE: In effect, this reduces overall share price by the combined fee
     # NOTE: may throw if Vault.totalAssets() > 1e64, or not called for more than a year
-    governance_fee: uint256 = (
+    governance_management_fee: uint256 = (
```
fubuloubu, line 1529: What if we just call this `management_fee`?

contracts/Vault.vy
```diff
@@ -1540,20 +1522,21 @@ def _reportLoss(strategy: address, loss: uint256):
...
-            * (block.timestamp - self.lastReport)
+            (self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets())
+            * (block.timestamp - self.strategies[strategy].lastReport)
             * self.managementFee
         )
         / MAX_BPS
         / SECS_PER_YEAR
     )
     strategist_fee: uint256 = 0  # Only applies in certain conditions
+    governance_performance_fee: uint256 = 0
```
fubuloubu, line 1539: `performance_fee`

fubuloubu (approved):

contracts/Vault.vy
Grandthrax: Sounds good. Done:
6298544c6229ffef262579cd9c90b8cbecb0c46a

contracts/Vault.vy
```diff
@@ -1563,13 +1546,17 @@ def _assessFees(strategy: address, gain: uint256):
...
     # NOTE: This must be called prior to taking new collateral,
     #       or the calculation will be wrong!
     # NOTE: This must be done at the same time, to ensure the relative
     #       ratio of governance_fee : strategist_fee is kept intact
-    total_fee: uint256 = governance_fee + strategist_fee
+    total_fee: uint256 = performance_fee + strategist_fee + management_fee
+    # ensure total_fee is not more than gain
+    if total_fee > gain:
+        total_fee = gain
+        management_fee = total_fee - performance_fee - strategist_fee
```
fubuloubu, line 1559: Same computed value, just makes it more clear what's going on here.
```suggestion
        management_fee = gain - performance_fee - strategist_fee
```

Also, this could be a DoS vuln on `report` if `performance_fee + strategist_fee > gain`, which shouldn't really happen in practice but is possible if `performanceFee + StrategyParams.performanceFee > 10_000 # BPS`

contracts/Vault.vy
```diff
@@ -1540,20 +1522,21 @@ def _reportLoss(strategy: address, loss: uint256):
...
-            * (block.timestamp - self.lastReport)
+            (self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets())
+            * (block.timestamp - self.strategies[strategy].lastReport)
             * self.managementFee
         )
         / MAX_BPS
         / SECS_PER_YEAR
     )
     strategist_fee: uint256 = 0  # Only applies in certain conditions
+    performance_fee: uint256 = 0
```
fubuloubu, line 1539: Just to clean up reading this section
```suggestion

    # Only applies in certain conditions
    strategist_fee: uint256 = 0
    performance_fee: uint256 = 0
```

fubuloubu (approved):
Needs a few tests showing the new behavior when `gain < total_fees`

contracts/Vault.vy
```diff
@@ -1558,6 +1558,7 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
     # ensure total_fee is not more than gain
     if total_fee > gain:
         total_fee = gain
+        # if combined management fee is greater than 100% then this will cause an underflow
```
fubuloubu, line 1561:
```suggestion
        # if total performance fee is greater than 100% then this will cause an underflow
```

tests/functional/strategy/test_fees.py
```diff
@@ -93,3 +93,29 @@ def test_delegated_fees(chain, rewards, vault, strategy, gov, token):
...
+    
+    # Performance fees set to standard 10%
+    vault.setPerformanceFee(1000, {"from": gov})
+    vault.updateStrategyPerformanceFee(strategy, 1000, {"from": gov})
+    chain.mine(timedelta=60 * 60 * 24 * 365)  # Mine a year at 2% mgmt fee
+    price_per_share_before = vault.pricePerShare()
+    strategy.harvest()
+
+    # Share price should not have changed because 100% of profit goes too fees. No more no less
+    assert vault.pricePerShare() == price_per_share_before
```
fubuloubu, line 121: Add something after this showing that 100% of the gain went towards fees

Grandthrax: done:
5e80fc6b121daaaac2a944de7658de1adcfab593

fubuloubu: I was thinking something more explicit e.g. `assert vault.strategies(strategy).dict()["totalGain"] == ...`

fubuloubu: But you can probably do something with before/after snapshots of shares for these two

fubuloubu (approved):

PR #271, opened 2021-04-12
feat: add management to initialize

steffenix:


scripts/deploy.py
```diff
@@ -119,6 +120,7 @@ def main():
             token,
             gov,
             rewards,
+            gov,
```
fubuloubu, line 123: Do we want the script to include this as an option?

fubuloubu (approved):

scripts/deploy.py
steffenix: Probably.

steffenix: updated on https://github.com/yearn/yearn-vaults/pull/271/commits/7e2de163b3bd9c8e49b6db8ee8ce733fadf2aed2

scripts/deploy.py
```diff
@@ -92,6 +92,7 @@ def main():
 
     rewards = get_address("Rewards contract", default="treasury.ychad.eth")
     guardian = get_address("Vault Guardian", default="dev.ychad.eth")
+    management = get_address("Vault Management", default="ychad.eth")
```
poolpitako, line 95: @fubuloubu @dougstorm we need an ens for sms

fubuloubu: `brain.ychad.eth`!

PR #273, opened 2021-04-14
feat: locked profit on deposit

Grandthrax:
1 - Make sure that we deposit and withdraw at same share price
2 - Add onreleased profit to next lockedprofit

PR #275, opened 2021-04-16
Individual profit degradation

### LINKED ISSUE #196: Track independent reports from strategies and aggregate total gain from them
Context:
https://github.com/iearn-finance/yearn-vaults/pull/167/files#r574205965

This is an optimization to improve on this solution to avoid strategies overwriting global lockedProfit value on reports

steffenix:
This is built on top of https://github.com/yearn/yearn-vaults/pull/273
fixes #196

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -13,6 +13,7 @@ struct StrategyParams {
     uint256 minDebtPerHarvest;
     uint256 maxDebtPerHarvest;
     uint256 lastReport;
+    uint256 lockedProfit;
     uint256 totalDebt;
     uint256 totalGain;
     uint256 totalLoss;
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -817,7 +817,6 @@ def _issueSharesForAmount(to: address, amount: uint256) -> uint256:
         # NOTE: if sqrt(token.totalSupply()) > 1e39, this could potentially revert
         precisionFactor: uint256 = self.precisionFactor
 
-        lockedFundsRatio: uint256 = (block.timestamp - self.lastReport) * self.lockedProfitDegration
         freeFunds: uint256 = self._totalAssets() - self._calculateLockedProfit()
         shares = precisionFactor * amount * totalSupply / freeFunds / precisionFactor
     else:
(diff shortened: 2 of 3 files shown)


contracts/BaseStrategy.sol
```diff
@@ -13,6 +13,7 @@ struct StrategyParams {
     uint256 minDebtPerHarvest;
     uint256 maxDebtPerHarvest;
     uint256 lastReport;
+    uint256 lockedProfit;
```
steffenix, line 16: That is the part that took me the longest to debug why harvest was withdrawing all the funds from the strategy, indeed lockedProfit=0 was used instead of totalDebt because I forgot to add this line here. Hopefully, interface will be possible in the future.

contracts/Vault.vy
```diff
@@ -201,6 +202,7 @@ event StrategyAddedToQueue:
 
 # NOTE: Track the total for overhead targeting purposes
 strategies: public(HashMap[address, StrategyParams])
+strategies_addresses: public(address[MAXIMUM_STRATEGIES])
```
fubuloubu, line 205: Instead of using an array, I was hoping it would be possible to update the aggregate metric `lockedProfit` based on some more advanced math, happening each time `report` is called.

I think #273 might actually be solving this with the change in `report` that takes what locked profit was and adds it to the new gains.

steffenix: The issue with that strategy is that some profit from a previous report will be locked for longer than it should.
If strategy A reports 300 USD gain it will be locked for 6 hours.
Then strategy B reports 150 USD gain 3 hours after A, 300 USD will be locked for another 6 hours.

What can happen if a lot of strategies keep on reporting gains and taking some of the previous report generated gains, as the price per share is considered without the locked funds, some will find opportunities to buy some shares when the locked value become interesting enough.

dougstorm:
Although this solution accounts for individual profit on each strategy, i think the cost/benefit of users increasing their gas cost on deposit / withdraw ten fold may not be justified TBH, is a minor issue to override the global lockedProfit compared to the N gas cost that users will pay for the main operations.

steffenix:
Agreed with @dougstorm

PR #277, opened 2021-04-19
fix: prevent shares from being deposited to wrong address

### LINKED ISSUE #276: Shares can be indirectly transferred to 0x0
Users can bypass transfer protections to send shares to the 0x0 address burning them.

Users can deposit funds to obtain shares in the Yearn vaults. The shares are ERC20-compatible and can be transferred freely, except to the address of the contract or the 0x0 address:

steffenix:
fixes #276

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -625,7 +625,7 @@ def _transfer(sender: address, receiver: address, amount: uint256):
     # See note on `transfer()`.
 
     # Protect people from accidentally sending their shares to bad places
-    assert not (receiver in [self, ZERO_ADDRESS])
+    assert receiver not in [self, ZERO_ADDRESS]
     self.balanceOf[sender] -= amount
     self.balanceOf[receiver] += amount
     log Transfer(sender, receiver, amount)
@@ -850,6 +850,7 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) -> 
     @return The issued Vault shares.
     """
     assert not self.emergencyShutdown  # Deposits are locked out
+    assert recipient not in [self, ZERO_ADDRESS]
 
     amount: uint256 = _amount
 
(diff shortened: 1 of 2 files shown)


contracts/Vault.vy
```diff
@@ -846,6 +846,7 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
     @return The issued Vault shares.
     """
     assert not self.emergencyShutdown  # Deposits are locked out
+    assert not (recipient in [self, ZERO_ADDRESS])
```
fubuloubu, line 849: Instead, add to `erc20_safe_transfer` and `erc20_safe_transferFrom` instead

banteg: `assert recipient not in [self, ZERO_ADDRESS]` would also work in latest Vyper

steffenix: `erc20_safe_transferFrom` is used here and will fail:
https://github.com/yearn/yearn-vaults/blob/master/contracts/Vault.vy#L1718

steffenix: @banteg https://github.com/yearn/yearn-vaults/pull/288

contracts/Vault.vy
```diff
@@ -854,8 +855,7 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
     if amount == MAX_UINT256:
         amount = min(
             self.depositLimit - self._totalAssets(),
-            self.token.balanceOf(msg.sender),
-        )
+            self.token.balanceOf(msg.sender))
```
fubuloubu, line 858: keep this the way it was

contracts/Vault.vy
```diff
@@ -855,7 +856,7 @@ def deposit(_amount: uint256 = MAX_UINT256, recipient: address = msg.sender) ->
         amount = min(
             self.depositLimit - self._totalAssets(),
             self.token.balanceOf(msg.sender),
-        )
+            )
```
fubuloubu, line 859: Undo this change, `black` would make it look like this

fubuloubu (approved):
Seems like the `hardhat.config.js` file belongs in another PR?

PR #279, opened 2021-04-19
fix: prevent setting wrong reward address

### LINKED ISSUE #278: Setting rewards address to 0x0 or self can block the fee computation
If governance tries to disable rewards by setting the address to receive them to 0x0 or the address of the contract, it can block the fee computation.

Users can deposit funds to obtain shares in the Yearn vaults. The shares are ERC20-compatible and can be transferred freely, except to the address of the contract or the 0x0 address:

steffenix:
fixes #278

### DIFF
--- a/tests/functional/vault/test_config.py
+++ b/tests/functional/vault/test_config.py
@@ -203,3 +203,6 @@ def test_vault_setLockedProfitDegration_range(gov, vault):
 def test_vault_setParams_bad_vals(gov, vault):
     with brownie.reverts():
         vault.setRewards(ZERO_ADDRESS, {"from": gov})
+
+    with brownie.reverts():
+        vault.setRewards(vault, {"from": gov})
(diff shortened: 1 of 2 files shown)


tests/functional/vault/test_config.py
```diff
@@ -197,3 +198,8 @@ def test_vault_setLockedProfitDegration_range(gov, vault):
     vault.setLockedProfitDegration(DEGREDATION_COEFFICIENT, {"from": gov})
     with brownie.reverts():
         vault.setLockedProfitDegration(DEGREDATION_COEFFICIENT + 1, {"from": gov})
+
+
+def test_vault_setParams_bad_vals(gov, vault):
+    with brownie.reverts():
+        vault.setRewards(ZERO_ADDRESS, {"from": gov})
```
fubuloubu, line 205:
```suggestion
    with brownie.reverts():
        vault.setRewards(ZERO_ADDRESS, {"from": gov})
        
    with brownie.reverts():
        vault.setRewards(vault, {"from": gov})
```

fubuloubu (approved):

PR #291, opened 2021-04-21
fix: vyper version

steffenix:


contracts/Registry.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.2.11
+# @version 0.2.12
```
fubuloubu, line 1: We can't bump this without redeploying the Registry. Unless there is a bug or issue we are vulnerable to with this contract, best to leave this off the table.

contracts/Vault.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.2.11
+# @version 0.2.12
```
fubuloubu, line 1: Must target an Vault release. Can you provide a summary of codesize and gas cost changes between 0.2.11 and 0.2.12 using our test suite?

PR #292, opened 2021-04-21
chore: bump brownie version

steffenix:


contracts/Registry.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.2.11
+# @version 0.2.12
```
fubuloubu, line 1: We can't bump this one unless we re-release the registry

contracts/Vault.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.2.11
+# @version 0.2.12
```
fubuloubu, line 1: Can't bump this one yet until a new Vaults version is ready

contracts/Registry.vy
steffenix: 👌

fubuloubu (approved):

PR #299, opened 2021-04-26
fix: avoid duplicates on withdrawal queue

steffenix:


.vscode/sftp.json
```diff
@@ -0,0 +1,9 @@
+{
```
fubuloubu, line 1: why is this file here?

fubuloubu (commented):
This was meant to be against `develop`?

contracts/Vault.vy
```diff
@@ -581,6 +581,13 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         if queue[i] == ZERO_ADDRESS and self.withdrawalQueue[i] == ZERO_ADDRESS:
             break
         assert self.strategies[queue[i]].activation > 0
+
+        for j in range(MAXIMUM_STRATEGIES):
+            if queue[j] == ZERO_ADDRESS and self.withdrawalQueue[j] == ZERO_ADDRESS:
+                break
+            if queue[i] == queue[j] and i != j:
+                raise "duplicates in queue"
```
fubuloubu, line 589: Would be cheaper to do this outside of the loop

fubuloubu (approved):

contracts/Vault.vy
steffenix: You mean having a second `for i loop` and inside a `jor j loop`?

fubuloubu: Yeah but removing the `self.withdrawalQueue[j]` line, which adds `2000 * n^2` gas, where `n` is the length of the queue.

steffenix: self.withdrawalQueue[j] == ZERO_ADDRESS is a mistake I have removed it.

contracts/Vault.vy
```diff
@@ -581,6 +581,14 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         if queue[i] == ZERO_ADDRESS and self.withdrawalQueue[i] == ZERO_ADDRESS:
             break
         assert self.strategies[queue[i]].activation > 0
+
+    for i in range(MAXIMUM_STRATEGIES):
+        for j in range(MAXIMUM_STRATEGIES):
+            if queue[j] == ZERO_ADDRESS:
+                break
+            if queue[i] == queue[j] and i != j:
+                raise "duplicates in queue"
```
fubuloubu, line 590:
```suggestion
            if i == j:
                continue
            assert queue[i] != queue[j]:
```

contracts/Vault.vy
```diff
@@ -581,6 +581,14 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         if queue[i] == ZERO_ADDRESS and self.withdrawalQueue[i] == ZERO_ADDRESS:
             break
         assert self.strategies[queue[i]].activation > 0
+
+    for i in range(MAXIMUM_STRATEGIES):
```
fubuloubu, line 585: I agree now that this can be combined with above to reduce overhead

steffenix: updated.

contracts/Vault.vy
```diff
@@ -581,6 +581,14 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         if queue[i] == ZERO_ADDRESS and self.withdrawalQueue[i] == ZERO_ADDRESS:
             break
         assert self.strategies[queue[i]].activation > 0
+
+        for j in range(MAXIMUM_STRATEGIES):
+            if queue[j] == ZERO_ADDRESS:
+                break
+            if i == j:
+                continue
+            assert queue[i] != queue[j]
```
fubuloubu, line 590: you might be able to get away with `for j in range(min(i, MAXIMUM_STRATEGIES)): ...` and avoid the redundant checks, or simple change the `if i == j: ...` condition to break instead of skip that loop

steffenix:
Made a mistake closing this

contracts/Vault.vy
steffenix: I have tried to do that with `for j in range(i - 1):` but that wasn't compiling

steffenix: Vyper version: 0.2.11
CompilerError: vyper returned the following errors:
```
Value must be a literal
  contract "contracts/Vault.vy", function "setWithdrawalQueue", line 585:8
       584
  ---> 585         for j in range(min(i, MAXIMUM_STRATEGIES)):
  -----------------^
       586             if queue[j] == ZERO_ADDRESS:
```

fubuloubu: Ah my bad. That was discussed, but never implemented. What I was thinking of was actually this:
```python
    for j in range(i, i + MAXIMUM_STRATEGIES):
        ...
```

Which... doesn't actually help us much in this instance (it allows "chunking up" a large iteration that might take multiple calls to fully work through without breaking the gas limit)

---

I think this would actually be the best:

```python
        # Make sure that `queue[i]` is not duplicated in the list
        for j in range(MAXIMUM_STRATEGIES):
            # NOTE: This will only check entries in `queue` that are `< i` (saves gas)
            if i == j:
                break
            assert queue[i] != queue[j]
```
Note that I removed the breaking condition  `queue[j] == ZERO_ADDRESS` because it shouldn't be possible to find a `j < i` that is an empty address given we already break out above if `queue[i]` is an empty address

contracts/Vault.vy
```diff
@@ -581,6 +581,14 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         if queue[i] == ZERO_ADDRESS and self.withdrawalQueue[i] == ZERO_ADDRESS:
             break
```
fubuloubu, line 582: Just noticed `self.withdrawalQueue[i] == ZERO_ADDRESS` is needed to break out of this loop.... Thinking about this condition, it doesn't seem to make sense and could be a weird corner case to this logic. I would suggest that instead it is modified to this:

```suggestion
        # old_queue[i] = self.withdrawalQueue[i]  # `old_queue` is defined before loop
        if queue[i] == ZERO_ADDRESS:
            # NOTE: Cannot use this method to remove entries from the queue
            assert old_queue[i] == ZERO_ADDRESS
            break
        # NOTE: Cannot use this method to add more entries to the queue
        assert old_queue[i] != ZERO_ADDRESS
```

---

Side note: we should have better testing around this logic to show that we can't add or remove entries from the queue using this reordering method. With the above suggested changes, we would have a check that entries are neither added nor removed, and that there are no duplicates. However, a forth condition is needed, and that is "all items in queue prior to reordering are still in queue afterwards".

steffenix: Got it to work with a check to make sure a new strategy isn't replacing an older one.

contracts/Vault.vy
```diff
@@ -550,11 +550,33 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+            if queue[j] == ZERO_ADDRESS:
+                alreadyExisting = True
+                break
+            if i == j:
+                continue
+            assert queue[i] != queue[j]
+            if old_queue[i] == queue[j]:
+                alreadyExisting = True
+            
+        assert alreadyExisting # dev: do not add non existing strategy
```
fubuloubu, line 578: Slight refactor for clarity
```suggestion
        alreadyExisting: bool = False
        for j in range(MAXIMUM_STRATEGIES):
            if old_queue[i] == queue[j]:
                # NOTE: Ensure that every entry in queue prior to reordering exists now
                alreadyExisting = True

            if j < i:
                # NOTE: This will only check for duplicate entries in queue after `i`
                continue
            assert queue[i] != queue[j]  # dev: do not add duplicate strategies

        assert alreadyExisting # dev: do not add new strategies
```

tests/functional/vault/test_strategies.py
```diff
@@ -346,6 +346,37 @@ def test_ordering(gov, vault, TestStrategy, rando):
...
+        )
+
+    # Do not remove strategies
+    with brownie.reverts():
+        vault.setWithdrawalQueue(
+            strategies[0:-2] + [ZERO_ADDRESS] * (20 - len(strategies[0:-2])),
+            {"from": gov},
+        )
+
+    other_strategy_list = strategies.copy()
```
fubuloubu, line 365:
```suggestion
    # Do not add new strategies
    other_strategy_list = strategies.copy()
```

fubuloubu (approved):

fubuloubu (approved):

contracts/Vault.vy
```diff
@@ -564,18 +564,18 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+        existsInOldQueue: bool = False
         for j in range(MAXIMUM_STRATEGIES):
-            if queue[j] == ZERO_ADDRESS:
-                alreadyExisting = True
-                break
-            if i == j:
+            if queue[i] == old_queue[j]:
+                # NOTE: Ensure that every entry in queue prior to reordering exists now
+                existsInOldQueue = True
+
```
steffenix, line 572: Was it intentional to remove the break here?

PR #312, opened 2021-05-03
fix: baseWrapper do not deposit dust 

### LINKED ISSUE #305: Fix occasional test case failure
Some CI runs are flaky because it occasionally triggers an assertion in the vaults code for `withdraw`
https://github.com/yearn/yearn-vaults/pull/302/checks?check_run_id=2457823927

image

Solution is to basically just make sure it doesn't withdraw dust

steffenix:
fixes #305

Doing 10 integration tests on: https://github.com/yearn/yearn-vaults/pull/313

### DIFF
--- a/contracts/BaseWrapper.sol
+++ b/contracts/BaseWrapper.sol
@@ -253,7 +253,7 @@ abstract contract BaseWrapper {
 
         // If we have extra, deposit back into `_bestVault` for `sender`
         // NOTE: Invariant is `withdrawn <= amount`
-        if (withdrawn > amount) {
+        if (withdrawn > amount && amount > _bestVault.pricePerShare().div(10**_bestVault.decimals())) {
             // Don't forget to approve the deposit
             if (token.allowance(address(this), address(_bestVault)) < withdrawn.sub(amount)) {
                 token.safeApprove(address(_bestVault), UNLIMITED_APPROVAL); // Vaults are trusted


steffenix:
I think it also needs https://github.com/yearn/yearn-vaults/pull/273/files#diff-b583ba5c4d7b67e56b9084eff9fe883e12a2cdd94a2ea25029322c6f3568d807R821 to be correct.

contracts/BaseWrapper.sol
```diff
@@ -253,7 +253,7 @@ abstract contract BaseWrapper {
 
         // If we have extra, deposit back into `_bestVault` for `sender`
         // NOTE: Invariant is `withdrawn <= amount`
-        if (withdrawn > amount) {
+        if (withdrawn > amount && amount > _bestVault.pricePerShare()) {
```
fubuloubu, line 256: The problem isn't that it's less than what's required for 1 share, it's that it's less than what's required for `1 / 10 ** decimals()` part of a share.

Also, this would keep the dust "trapped" in whatever implements `BaseWrapper`

steffenix: I don't see how it will be kept as dust, https://github.com/steffenix/yearn-vaults/blob/2960055466d5ab68553cfa03f34f1674bb57fd98/contracts/BaseWrapper.sol#L263 won't be executed and the overall withdrawn will be sent back to the receiver

fubuloubu: > Yes indeed, should we expose `_shareValue`?

I think so

fubuloubu: > I don't see how it will be kept as dust, https://github.com/steffenix/yearn-vaults/blob/2960055466d5ab68553cfa03f34f1674bb57fd98/contracts/BaseWrapper.sol#L263 won't be executed and the overall withdrawn will be sent back to the receiver

Fair point.

fubuloubu (approved):

PR #314, opened 2021-05-04
fix: comments in Vault.vy 

antoncoding:
Fix comments in `Vault.vy`

1. The default symbol starts with `yv` instead of `y`
2. Gas of 2 SLOADs drops from 4000 to 200.


steffenix:
Make sure your commit history contains messages that comply with https://github.com/conventional-changelog/commitlint/#what-is-commitlint

steffenix (approved):

PR #326, opened 2021-05-05
fix: make sure _assessFees isn't called twice within the same block

### LINKED ISSUE #322: Management fee can be avoided
Block `report` from being called if time since last report is 0 (same block)

steffenix:
fixes #322

### DIFF
--- a/tests/functional/strategy/test_migration.py
+++ b/tests/functional/strategy/test_migration.py
@@ -4,9 +4,10 @@ from brownie import ZERO_ADDRESS
 
 
 def test_good_migration(
-    token, strategy, vault, gov, strategist, guardian, TestStrategy, rando
+    token, strategy, vault, gov, strategist, guardian, TestStrategy, rando, chain
 ):
     # Call this once to seed the strategy with debt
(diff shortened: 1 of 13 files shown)


tests/functional/strategy/test_migration.py
```diff
@@ -4,9 +4,11 @@
 
 
 def test_good_migration(
-    token, strategy, vault, gov, strategist, guardian, TestStrategy, rando
+    token, strategy, vault, gov, strategist, guardian, TestStrategy, rando, chain
 ):
     # Call this once to seed the strategy with debt
+    chain.sleep(1)
+    chain.sleep(1)
```
fubuloubu, line 11: Why is this here twice?

fubuloubu (approved):

zgfzgf:
assert duration != 0 should be in report function

steffenix:
> assert duration != 0 should be in report function

It's more gas efficient to have it within `_assessFees` for strategies that do not misbehave.

zgfzgf:
I see. thank you

PR #327, opened 2021-05-10
fix compute credit error

zgfzgf:
self.totalDebt -= debtPayment
should  tranfer token from msg.send to self

so put self._creditAvailable before


steffenix:
Please fix the commit message and PR title

Issue #333, opened 2021-05-11
PrecisionFactor is useless


zgfzgf:
https://github.com/yearn/yearn-vaults/blob/3db9deded94c11b0536214be3fc1ac57b439212c/contracts/Vault.vy#L806

for example: precisionFactor=10 
amount * totalSupply / self._totalAssets()=23.45 float 

not use  precisionFactor
  The results are **23**

use  precisionFactor
precisionFactor*amount * totalSupply / self._totalAssets()=234.5  is 234
234/precisionFactor=234/10=**23**

so precisionFactor is useless

steffenix:
For that particular line, it is useless yes.
On more complex calculations It remains relevant.

zgfzgf:
maybe self.precisionFactor is useless

PR #335, opened 2021-05-11
fix: precision factor is useless

steffenix:
As @zgfzgf pointed out on #333 the precision factor isn't improving precision most of the time.
It would only be useful if we were doing a multiplication after a division.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -73,7 +73,6 @@ event Approval:
 name: public(String[64])
 symbol: public(String[32])
 decimals: public(uint256)
-precisionFactor: public(uint256)
 
 balanceOf: public(HashMap[address, uint256])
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):
I think there was one or two cases where this extra precision factor did in fact help. Regardless, it would be nice to have an analysis that shows more completely that it is not doing the job it is meant to do.

Little background: this was added in a previous audit to account for tokens with decimals like GUSD, where 1 token = $0.01, so precision loss could quickly add up.

steffenix:
I have reviewed all our calculations following the comment from @zgfzgf and it can only be useful if we multiple after a division which we do not do.

steffenix:
> do we have a test that was targeting the precision factor logic specifically? if not can we add it to make sure there's no regression here?

We have tests running with two decimals tokens.
No test that are specific to precision.

PR #307, opened 2021-04-30
fix: divide by 0 error in _reportLoss (#306)

fp-crypto:
This PR fixes a divide by 0 error that manifests when a vault has a total loss and `harvest` is called. The divide by 0 is only triggered when there is a total loss. Included is a new test that demonstrates the issue when run against the existing `master` branch but is fixed in this PR.

fixes #306

PR #337, opened 2021-05-12
fix: prevent division per zero

### LINKED ISSUE #306: Div by zero in Vault code
To repro:

 - Create a vault
 - Deploy a strategy
 - Deposit a token
 - Make the strategy send ALL tokens to `ZERO_ADDRESS`
 - Call `harvest()`

`_reportLoss()` will revert because `self._totalAssets()` is 0.

Reported by @fp-crypto!

### LINKED ISSUE #316: Ensure that this value is not user-manipulatable.
Suggested change to ensure that this value is not user-manipulatable in `_reportLoss`

```python
  ratio_change: uint256 = min(
        # NOTE: This calculation isn't 100% precise, the adjustment is ~10%-20% more severe due to EVM math
        precisionFactor * MAX_BPS * loss / self.totalDebt * MAX_BPS / self.debtRatio / precisionFactor,
        self.strategies[strategy].debtRatio,
   )
```

Reference: #307

steffenix:
closes #316 & closes #306

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -960,7 +960,7 @@ def _reportLoss(strategy: address, loss: uint256):
     if self.debtRatio != 0: # if vault with single strategy that is set to EmergencyOne
         ratio_change = min(
             # NOTE: This calculation isn't 100% precise, the adjustment is ~10%-20% more severe due to EVM math
-            MAX_BPS * loss / (self.totalDebt * MAX_BPS / self.debtRatio),
+            loss * self.debtRatio / self.totalDebt,
             self.strategies[strategy].debtRatio,
     )
     # Finally, adjust our strategy's parameters by the loss
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -954,13 +954,18 @@ def _reportLoss(strategy: address, loss: uint256):
     # Loss can only be up the amount of debt issued to strategy
     totalDebt: uint256 = self.strategies[strategy].totalDebt
     assert totalDebt >= loss
+
+    # Also, make sure we reduce our trust with the strategy by the amount of loss
+    precisionFactor: uint256 = self.precisionFactor
+    ratio_change: uint256 = min(
+        precisionFactor * MAX_BPS * loss / self._totalAssets() / precisionFactor,
+        self.strategies[strategy].debtRatio,
+    )
+ 
+    # Finally, adjust our strategy's parameters by the loss
     self.strategies[strategy].totalLoss += loss
     self.strategies[strategy].totalDebt = totalDebt - loss
     self.totalDebt -= loss
-
-    # Also, make sure we reduce our trust with the strategy by the same amount
-    debtRatio: uint256 = self.strategies[strategy].debtRatio
-    ratio_change: uint256 = min(loss * MAX_BPS / self._totalAssets(), debtRatio)
     self.strategies[strategy].debtRatio -= ratio_change
     self.debtRatio -= ratio_change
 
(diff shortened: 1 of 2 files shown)


steffenix:
https://github.com/yearn/yearn-vaults/pull/307#pullrequestreview-649805790

If the `self.totalDebt -= loss` isn't done before we calculate the ratio_change then the ratio_change change is calculated on the amount before the loss. 

On the tests we have:

- Start: vault = 4500 strategy=500
- Strategy loss 100: vault = 4500 strategy=400 but totalAssets = 5000
- We add back 100 to the vault. vault = 4600 strategy=400 but totalAssets = 5100
- without changes: strategy report -> ratio_change = 10_000 * 100 / 5000 = 200
- with the changes: strategy report -> ratio_change = 10_000 * 100 / 5100 = 196

With these changes, the strategy that has losses will keep a higher debt ratio.
Is that what you expect @fubuloubu

fubuloubu:
> Is that what you expect with the changes you suggested @fubuloubu ?

No. In your example, 20% of the strategy funds were lost. The strategy has 10% of the total funds. Therefore a 2% adjustment (10% of 20%) should be made to its ratio.

However in my suggested change in https://github.com/yearn/yearn-vaults/issues/316#issue-875563254, the value would be `10_000 * 100 / (500 * 10_000 / 1_000)` as the strategy total debt would be normalized using the current debt ratio to whatever the total should be considered to be prior to any updates.

I thibk these two fixes should be worked together.

fubuloubu:
Also van you add @fp-crypto as a co-author on the first commit?

steffenix:
> Also van you add @fp-crypto as a co-author on the first commit?

Need an email to do so `Co-authored-by: name <name@example.com>`

steffenix:
@fubuloubu updated with changes needed for #316

contracts/Vault.vy
```diff
@@ -1438,12 +1443,14 @@ def removeStrategyFromQueue(strategy: address):
 @internal
 def _debtOutstanding(strategy: address) -> uint256:
     # See note on `debtOutstanding()`.
+    if self.debtRatio == 0:
+        return 0
```
fubuloubu, line 1447: This would be better, in case there's a scenario where we set it all to 0 but there's still outstanding debt
```suggestion
        return strategy_totalDebt
```

fubuloubu (approved):

steffenix:
With a vault with a single strategy and that strategy is to EmergencyExit mode we have a division by zero.

contracts/Vault.vy
```diff
@@ -954,13 +954,19 @@ def _reportLoss(strategy: address, loss: uint256):
     # Loss can only be up the amount of debt issued to strategy
     totalDebt: uint256 = self.strategies[strategy].totalDebt
     assert totalDebt >= loss
+
+    # Also, make sure we reduce our trust with the strategy by the amount of loss
+    ratio_change: uint256 = 0
+    if self.debtRatio != 0: # if vault with single strategy that is set to EmergencyOne
+        ratio_change = min(
+            # NOTE: This calculation isn't 100% precise, the adjustment is ~10%-20% more severe due to EVM math
+            MAX_BPS * loss / (self.totalDebt * MAX_BPS / self.debtRatio),
```
fubuloubu, line 963: We can actually cancel both uses of `MAX_BPS`, and rearrange terms as follows
```suggestion
            loss * self.debtRatio / self.totalDebt,
```

NOTE: use "reciprocal of reciprocal" rule

fubuloubu (approved):

PR #340, opened 2021-05-13
fix: delete userless implements

zgfzgf:
maybe use proxy contract.  implements is use


fubuloubu:
`implements: <InterfaceType>` in Vyper is syntactic sugar around a compiler check that all methods described in `InterfaveType` are implemented. This does no harm to leave in.

zgfzgf:
Yes

fubuloubu:
Meant to close with my comment above

PR #343, opened 2021-05-14
perf: optimize code

zgfzgf:
optimize code
and corresponding to _creditAvailable https://github.com/yearn/yearn-vaults/blob/f7a6bf53e133c9a39d867771c1a7d442b6488710/contracts/Vault.vy#L1471-L1474

It's easier to understand

PR #344, opened 2021-05-14
perf: optimize code in func _debtOutstanding

zgfzgf:
#343 
optimize code
and corresponding to _creditAvailable https://github.com/yearn/yearn-vaults/blob/f7a6bf53e133c9a39d867771c1a7d442b6488710/contracts/Vault.vy#L1471-L1474

It's easier to understand


zgfzgf:
why error?

fubuloubu:
> why error?

Theres an inconsistency in our fuzzer case that sometimes gets triggered

fubuloubu (approved):

PR #345, opened 2021-05-17
perf: assess fees develop

zgfzgf:
prevent overflow
probable    self.performanceFee + self.strategies[strategy].performanceFee >= MAX_BPS 
https://github.com/yearn/yearn-vaults/blob/eb575fd1bba029aab86227ffae5a4c0bdeb37957/contracts/Vault.vy#L451-L463


fubuloubu (changes requested):
This method also assesses the management fee, which is being skipped by this update

zgfzgf:
close it, merge main. management fee is not update now

PR #347, opened 2021-05-17
fix: performanceFee error

zgfzgf:
before OK
merge develop to main, it is wrong.

_assessFees func
gain - performance_fee - strategist_fee


steffenix:
That's not going to be consistent with the rest.
The code on your PR is what we use to have but if you look at the code there are several places to change strategy and vault fees.
If you update performance_fee calling `setPerformanceFee` you would have to loop over the strategies (which is not possible for now we are using a hashmap) and check that each strategy fee + the new performance_fee isn't over 100% this is not gas efficient.
The workaround is to make sure performance_fee and strategist_fee aren't more than 50%

honeywest:
I see

PR #353, opened 2021-05-18
refactor: move condition shares calculation and save gas

t4sk:
https://github.com/yearn/yearn-vaults/blob/master/contracts/Vault.vy#L1044
https://github.com/yearn/yearn-vaults/blob/master/contracts/Vault.vy#L1089-L1090

this condition does not change unless tokens are withdrawn from strategies, so move the code  to where it can possibly be true.


steffenix:
@fubuloubu that seems right.

fubuloubu (approved):

PR #348, opened 2021-05-17
fix: performanceFee error

zgfzgf:
#347

PR #354, opened 2021-05-18
perf: _assessFees

zgfzgf:
#345 
prevent overflow
probable    self.performanceFee + self.strategies[strategy].performanceFee >= MAX_BPS 
https://github.com/yearn/yearn-vaults/blob/eb575fd1bba029aab86227ffae5a4c0bdeb37957/contracts/Vault.vy#L451-L463

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1584,7 +1584,7 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
         / SECS_PER_YEAR
     )
 
-    self. managementDebt +=  management_fee
+    self.managementDebt +=  management_fee
 
     if gain == 0:
         return 0
@@ -1607,11 +1607,11 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
     #       or the calculation will be wrong!
     # NOTE: This must be done at the same time, to ensure the relative
     #       ratio of governance_fee : strategist_fee is kept intact
-    total_fee: uint256 = performance_fee + strategist_fee + self. managementDebt
-    self. managementDebt = 0
+    total_fee: uint256 = performance_fee + strategist_fee + self.managementDebt
+    self.managementDebt = 0
     # ensure total_fee is not more than gain
     if total_fee > gain:
-        self. managementDebt = total_fee - gain
+        self.managementDebt = total_fee - gain
         total_fee = gain
     if total_fee > 0:  # NOTE: If mgmt fee is 0% and no gains were realized, skip
         reward: uint256 = self._issueSharesForAmount(self, total_fee)
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1575,6 +1575,8 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
     if gain == 0:
         return 0
 
+    assert self.strategies[strategy].performanceFee + self.performanceFee <= MAX_BPS
+
     management_fee: uint256 = (
         (
             (self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets())
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -215,6 +215,7 @@ lockedProfitDegradation: public(uint256) # rate per block of degradation. DEGRAD
 rewards: public(address)  # Rewards contract where Governance fees are sent to
 # Governance Fee for management of Vault (given to `rewards`)
 managementFee: public(uint256)
+managementDebt: public(uint256)
 # Governance Fee for performance of Vault (given to `rewards`)
 performanceFee: public(uint256)
 MAX_BPS: constant(uint256) = 10_000  # 100%, or 10k basis points
@@ -290,6 +291,7 @@ def initialize(
     self.performanceFee = 1000  # 10% of yield (per Strategy)
     log UpdatePerformanceFee(convert(1000, uint256))
     self.managementFee = 200  # 2% per year
+    self.managementDebt = 0
     log UpdateManagementFee(convert(200, uint256))
     self.lastReport = block.timestamp
     self.activation = block.timestamp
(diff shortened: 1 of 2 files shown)


zgfzgf:
add var managementDebt 
in order to  management fee update.

zgfzgf:
I don't know where the test  wrong

contracts/Vault.vy
```diff
@@ -290,6 +291,7 @@ def initialize(
     self.performanceFee = 1000  # 10% of yield (per Strategy)
     log UpdatePerformanceFee(convert(1000, uint256))
     self.managementFee = 200  # 2% per year
+    self.managementDebt = 0
```
steffenix, line 294: Do not change default management fee

contracts/Vault.vy
```diff
@@ -1582,32 +1584,35 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
...
-    # Only applies in certain conditions
-    strategist_fee: uint256 = 0
-    performance_fee: uint256 = 0
+    self.managementDebt +=  management_fee
+
+    if gain == 0:
+        return 0
+
+    # if total performance fee is greater than 100% then this will cause an underflow
+    assert self.strategies[strategy].performanceFee + self.performanceFee <= MAX_BPS
```
steffenix, line 1593: It won't as we going to merge this
https://github.com/yearn/yearn-vaults/pull/348

contracts/Vault.vy
zgfzgf: in order to keep the managementFee. 
If it's not enough this time, it will be deducted next time

contracts/Vault.vy
```diff
@@ -1582,32 +1584,35 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
         / SECS_PER_YEAR
     )
 
-    # Only applies in certain conditions
-    strategist_fee: uint256 = 0
-    performance_fee: uint256 = 0
+    self.managementDebt +=  management_fee
+
+    if gain == 0:
+        return 0
```
steffenix, line 1590: Fee should be at least `management_fee`

zgfzgf: when gain =0; this time management_fee keep.
it will be deducted next time

contracts/Vault.vy
zgfzgf: in order to safe. so make more judgments

steffenix: You have to keep it gas efficient the other checks are covering it

zgfzgf: I see.
now delete it. test may fail. may update test

contracts/Vault.vy
steffenix: I am not sure Debt is an appropriate word as management doesn't need to pay back, whereas strategy Debt is what the strategy needs to pay back

contracts/Vault.vy
steffenix: management_fee can be assets without gains, it's not based on strategy performance but vault managed assets.

contracts/Vault.vy
```diff
@@ -1582,32 +1584,35 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
...
     #       ratio of governance_fee : strategist_fee is kept intact
-    total_fee: uint256 = performance_fee + strategist_fee + management_fee
+    total_fee: uint256 = performance_fee + strategist_fee + self.managementDebt
+    self.managementDebt = 0
     # ensure total_fee is not more than gain
     if total_fee > gain:
+        self.managementDebt = total_fee - gain
         total_fee = gain
-        # if total performance fee is greater than 100% then this will cause an underflow
-        management_fee = gain - performance_fee - strategist_fee
```
steffenix, line 1610: management_fee is indeed not used after being updated @fubuloubu @Grandthrax 

https://github.com/yearn/yearn-vaults/pull/263/files#diff-b583ba5c4d7b67e56b9084eff9fe883e12a2cdd94a2ea25029322c6f3568d807R1562

contracts/Vault.vy
steffenix: > when gain =0; this time management_fee keep.
> it will be deducted next time

Please add a test that test managementDebt behavior when gains are 0.

contracts/Vault.vy
fubuloubu: This is because the total balance is sent to the rewards for the management fee. So indeed, this invariant is kept but it is unused because the resulting value sent to rewards is:
```
_issueSharesForAmount( management_fee + performance_fee )
= _issueSharesForAmount( (gain - performance_fee - strategist_fee) + performance_fee )
= _issueSharesForAmount( gain - strategist_fee )
= token.balanceOf(self)
```

fubuloubu (commented):
I'm having a hard time understanding what this PR is for

Usually, I don't like to add new storage variables without a good reason why they are needed

contracts/Vault.vy
zgfzgf: https://github.com/yearn/yearn-vaults/blob/cc691c4d725b593b5a9ee406fd6dbe9509435aa8/contracts/Vault.vy#L1605-L1609
when total_fee < gain
exec management_fee may < management_fee
should be add var storage management_fee - (exec management_fee)
https://github.com/yearn/yearn-vaults/blob/cc691c4d725b593b5a9ee406fd6dbe9509435aa8/contracts/Vault.vy#L1575-L1583

fubuloubu: Split this into two PRs:
1. The above 2 lines removed, with no other changes (with the documented reasoning that there is no need for this variable do to the implicit invariant that all leftover minted shares are sent to `rewards`)
2. This new "fix" (which we are having trouble parsing) using `managementDebt` and the associated updates to Vault and test cases

fubuloubu (changes requested):

zgfzgf:
OK, close it. open two new PR

PR #355, opened 2021-05-18
chore: bump vyper version

steffenix:


fubuloubu:
Do we want to update the 0.4.0 tag for this?

fubuloubu (approved):

contracts/Registry.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.2.11
+# @version 0.2.12
```
fubuloubu, line 1: Can't update the Registry

fubuloubu (changes requested):

steffenix:
> Do we want to update the 0.4.0 tag for this?

I am thinking we should tag 0.4 with that

fubuloubu:
> > Do we want to update the 0.4.0 tag for this?
> 
> I am thinking we should tag 0.4 with that

Let's do it!

Just not the Registry lol

steffenix:
> > > Do we want to update the 0.4.0 tag for this?
> > 
> > 
> > I am thinking we should tag 0.4 with that
> 
> Let's do it!
> 
> Just not the Registry lol

What's about Registry? Just don't need to redeploy it or something will breek?

fubuloubu:
> What's about Registry? Just don't need to redeploy it or something will breek?

Upgrading the Registry requires replaying all of the events in the previous version of the registry, as well as coordinating upgrades across all AffiliateWrappers and yTokens (which aren't deployed yet)

steffenix:
I do not understand why we can't bump the Registry version we just don't deploy it again.

fubuloubu:
> I do not understand why we can't bump the Registry version we just don't deploy it again.

We can deploy it again, its just the deployment process is much more involved than with a new `Vault` API version. There are many downstrram users of the registry that would need to be updated, and a large amount of transactions that need to be replayed on the newly deployed registry (20+ transactions)

Its possible, but not very simple, and not really worth it just to upgrade the compiler version, unless a critical bug were found or there was a new feature we needed.

steffenix:
@fubuloubu I have made the update

fubuloubu (approved):

PR #363, opened 2021-05-20
fix: storage management debt

zgfzgf:
#354
add var managementDebt

step 1: implementation
step 2: fix old tests
step 3: add new tests

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -215,6 +215,7 @@ lockedProfitDegradation: public(uint256) # rate per block of degradation. DEGRAD
 rewards: public(address)  # Rewards contract where Governance fees are sent to
 # Governance Fee for management of Vault (given to `rewards`)
 managementFee: public(uint256)
+managementDebt: public(uint256)
 # Governance Fee for performance of Vault (given to `rewards`)
 performanceFee: public(uint256)
 MAX_BPS: constant(uint256) = 10_000  # 100%, or 10k basis points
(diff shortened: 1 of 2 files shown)


zgfzgf:
I don't know why it's wrong

steffenix:
management_fee can be assets without gains, it's not based on strategy performance but vault managed assets.

What is the idea behind these changes?

Like @fubuloubu said `Usually, I don't like to add new storage variables without a good reason why they are needed`

zgfzgf:
block 1000, block 2000, block 3000  harvest
may block 2000 harvest. active exec management fee < receivable management fee.   debt=receivable - exec
block 3000 may pay debt

steffenix:
> block 1000, block 2000, block 3000 harvest
> may block 2000 harvest. active exec management fee < receivable management fee. debt=receivable - exec
> block 3000 may pay debt

Most likely it will never be the case, management Fee is 2% annualized and assigned to Treasury. If that ever becomes an issue that vault management fee is at loss we will reconsider.

fubuloubu:
> I don't understand what your needs are?

Yes, I think there is some miscommunication, I am not sure what the issue you were trying to solve is.

zgfzgf:
> > I don't understand what your needs are?
> 
> Yes, I think there is some miscommunication, I am not sure what the issue you were trying to solve is.

first, I think #366 #345 is OK. you said:"This method also assesses the management fee, which is being skipped by this update"
so add var managementDebt

PR #366, opened 2021-05-21
perf: assessFees when gain==0

zgfzgf:
It's almost the same as #345 
@fubuloubu  in #345 you said:"This method also assesses the management fee, which is being skipped by this update"
as in #363 add var managementDebt

If there is no need to retain management fee, the PR is OK.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1579,6 +1579,7 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
     assert duration != 0 # can't assessFees twice within the same block
 
     if gain == 0:
+        # NOTE: The fees are not charged if there hasn't been any gains reported
         return 0
 
     management_fee: uint256 = (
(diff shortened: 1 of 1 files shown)


contracts/Vault.vy
```diff
@@ -1578,6 +1578,9 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
     duration: uint256 = block.timestamp - self.strategies[strategy].lastReport
     assert duration != 0 # can't assessFees twice within the same block
 
+    if gain == 0:
+        return 0
```
fubuloubu, line 1582:
```suggestion
        # NOTE: The fees are not charged if there hasn't been any gains reported
        return 0
```

fubuloubu (approved):
Add that comment, and we are good to go

contracts/Vault.vy
zgfzgf: OK

PR #367, opened 2021-05-21
style: update code style

zgfzgf:
like python. style is important


fubuloubu (approved):
Yeah, for a bit we tried to make a linter for Vyper, but that fell away

check it out here: https://github.com/vyperlang/blackadder

PR #373, opened 2021-05-25
fix: check debtRatio

zgfzgf:


steffenix:
That won't happen `_revokeStrategy` set the migrated strategy to 0.

steffenix:
`_revokeStrategy` set the migrated strategy to 0, the MAX_BPS and strategy debtRatio is checked with `addStrategy` and `updateStrategyDebtRatio`

PR #374, opened 2021-05-25
feat: add price update checks on report

steffenix:
This is my first suggestion to prevent strategies from reporting a loss that could have big price changes.

I will also add some changes to BaseStrategy on another PR.

More details on: https://hackmd.io/MCg207ecTj-ZtUgnU-dGzw

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -16,7 +16,7 @@ struct StrategyParams {
     uint256 totalDebt;
     uint256 totalGain;
     uint256 totalLoss;
-    uint256 ratioChangeLimit;
+    uint256 changeLimitRatio;
     bool enforeChangeLimit;
 }
 
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -16,8 +16,6 @@ struct StrategyParams {
     uint256 totalDebt;
     uint256 totalGain;
     uint256 totalLoss;
-    uint256 changeLimitRatio;
-    bool enforeChangeLimit;
 }
 
 interface VaultAPI is IERC20 {
@@ -181,43 +179,16 @@ abstract contract BaseStrategy {
     using SafeMath for uint256;
     using SafeERC20 for IERC20;
     string public metadataURI;
+    uint256 constant MAX_BPS = 10_000;
+    uint256 public changeLimitRatio = 300;
+    bool public enforeChangeLimit;
 
-    /**
-     * @notice
-     *  Used to track which version of `StrategyAPI` this Strategy
-     *  implements.
-     * @dev The Strategy's version must match the Vault's `API_VERSION`.
-     * @return A string which holds the current API version of this contract.
-     */
     function apiVersion() public pure returns (string memory) {
         return "0.4.0";
     }
 
-    /**
-     * @notice This Strategy's name.
-     * @dev
-     *  You can use this field to manage the "version" of this Strategy, e.g.
-     *  `StrategySomethingOrOtherV1`. However, "API Version" is managed by
-     *  `apiVersion()` function above.
-     * @return This Strategy's name.
-     */
     function name() external virtual view returns (string memory);
 
-    /**
-     * @notice
-     *  The amount (priced in want) of the total assets managed by this strategy should not count
-     *  towards Yearn's TVL calculations.
-     * @dev
-     *  You can override this field to set it to a non-zero value if some of the assets of this
-     *  Strategy is somehow delegated inside another part of of Yearn's ecosystem e.g. another Vault.
-     *  Note that this value must be strictly less than or equal to the amount provided by
-     *  `estimatedTotalAssets()` below, as the TVL calc will be total assets minus delegated assets.
-     *  Also note that this value is used to determine the total assets under management by this
-     *  strategy, for the purposes of computing the management fee in `Vault`
-     * @return
-     *  The amount of assets this strategy manages that should not be included in Yearn's Total Value
-     *  Locked (TVL) calculation across it's ecosystem.
-     */
     function delegatedAssets() external virtual view returns (uint256) {
         return 0;
     }
@@ -229,7 +200,6 @@ abstract contract BaseStrategy {
 
     IERC20 public want;
 
-    // So indexers can keep track of this
     event Harvested(uint256 profit, uint256 loss, uint256 debtPayment, uint256 debtOutstanding);
 
     event UpdatedStrategist(address newStrategist);
@@ -250,26 +220,16 @@ abstract contract BaseStrategy {
 
     event UpdatedMetadataURI(string metadataURI);
 
-    // The minimum number of seconds between harvest calls. See
-    // `setMinReportDelay()` for more details.
     uint256 public minReportDelay;
 
-    // The maximum number of seconds between harvest calls. See
-    // `setMaxReportDelay()` for more details.
     uint256 public maxReportDelay;
 
-    // The minimum multiple that `callCost` must be above the credit/profit to
-    // be "justifiable". See `setProfitFactor()` for more details.
     uint256 public profitFactor;
 
-    // Use this to adjust the threshold at which running a debt causes a
-    // harvest trigger. See `setDebtThreshold()` for more details.
     uint256 public debtThreshold;
 
-    // See note on `setEmergencyExit()`.
     bool public emergencyExit;
 
-    // modifiers
     modifier onlyAuthorized() {
         require(msg.sender == strategist || msg.sender == governance(), "!authorized");
         _;
@@ -306,85 +266,55 @@ abstract contract BaseStrategy {
     }
 
     constructor(address _vault) public {
-        _initialize(_vault, msg.sender, msg.sender, msg.sender);
+        _initialize(_vault, msg.sender, msg.sender, msg.sender, 300);
     }
 
-    /**
-     * @notice
-     *  Initializes the Strategy, this is called only once, when the
-     *  contract is deployed.
-     * @dev `_vault` should implement `VaultAPI`.
-     * @param _vault The address of the Vault responsible for this Strategy.
-     * @param _strategist The address to assign as `strategist`.
-     * The strategist is able to change the reward address
-     * @param _rewards  The address to use for pulling rewards.
-     * @param _keeper The adddress of the _keeper. _keeper
-     * can harvest and tend a strategy.
-     */
     function _initialize(
         address _vault,
         address _strategist,
         address _rewards,
-        address _keeper
+        address _keeper,
+        uint256 _changeLimitRatio
     ) internal {
         require(address(want) == address(0), "Strategy already initialized");
 
         vault = VaultAPI(_vault);
         want = IERC20(vault.token());
-        want.safeApprove(_vault, uint256(-1)); // Give Vault unlimited access (might save gas)
+        want.safeApprove(_vault, uint256(-1));
         strategist = _strategist;
         rewards = _rewards;
         keeper = _keeper;
+        changeLimitRatio = changeLimitRatio;
 
-        // initialize variables
         minReportDelay = 0;
         maxReportDelay = 86400;
         profitFactor = 100;
         debtThreshold = 0;
 
-        vault.approve(rewards, uint256(-1)); // Allow rewards to be pulled
+        vault.approve(rewards, uint256(-1));
     }
 
-    /**
-     * @notice
-     *  Used to change `strategist`.
-     *
-     *  This may only be called by governance or the existing strategist.
-     * @param _strategist The new address to assign as `strategist`.
-     */
     function setStrategist(address _strategist) external onlyAuthorized {
         require(_strategist != address(0));
         strategist = _strategist;
         emit UpdatedStrategist(_strategist);
     }
 
-    /**
-     * @notice
-     *  Used to change `keeper`.
-     *
-     *  `keeper` is the only address that may call `tend()` or `harvest()`,
-     *  other than `governance()` or `strategist`. However, unlike
-     *  `governance()` or `strategist`, `keeper` may *only* call `tend()`
-     *  and `harvest()`, and no other authorized functions, following the
-     *  principle of least privilege.
-     *
-     *  This may only be called by governance or the strategist.
-     * @param _keeper The new address to assign as `keeper`.
-     */
     function setKeeper(address _keeper) external onlyAuthorized {
         require(_keeper != address(0));
         keeper = _keeper;
         emit UpdatedKeeper(_keeper);
     }
 
-    /**
-     * @notice
-     *  Used to change `rewards`. EOA or smart contract which has the permission
-     *  to pull rewards from the vault.
-     *
-     *  This may only be called by the strategist.
-     * @param _rewards The address to use for pulling rewards.
-     */
+    function setStrategyEnforeChangeLimit(bool _enforeChangeLimit) external onlyAuthorized {
+        enforeChangeLimit = _enforeChangeLimit;
+    }
+
+    function setStrategychangeLimitRatio(uint256 _changeLimitRatio) external onlyAuthorized {
+        require(_changeLimitRatio < MAX_BPS);
+        changeLimitRatio = _changeLimitRatio;
+    }
+
     function setRewards(address _rewards) external onlyStrategist {
         require(_rewards != address(0));
         vault.approve(rewards, 0);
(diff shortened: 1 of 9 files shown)


contracts/Vault.vy
```diff
@@ -1664,16 +1682,23 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
         Amount Strategy has made available to cover outstanding debt
     @return Amount of debt outstanding (if totalDebt > debtLimit or emergency shutdown).
     """
+    strategy: StrategyParams = self.strategies[msg.sender] # should only be used for reading
```
fubuloubu, line 1685: This actually loads the entire struct from storage, so if youre only using a few items from it, better to load them individually

contracts/BaseStrategy.sol
```diff
@@ -16,6 +16,8 @@ struct StrategyParams {
     uint256 totalDebt;
     uint256 totalGain;
     uint256 totalLoss;
+    uint256 healtyChanges;
```
fubuloubu, line 19: Lets call this `lossRatioLimit`

contracts/BaseStrategy.sol
```diff
@@ -16,6 +16,8 @@ struct StrategyParams {
...
+    bool doHealthCheck;
```
fubuloubu, line 20: Maybe `checkLoss`

contracts/Vault.vy
```diff
@@ -1664,16 +1682,23 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
...
 
     # We have a loss to report, do it before the rest of the calculations
     if loss > 0:
+        pricePerShare: uint256 = self._shareValue(10 ** self.decimals)
         self._reportLoss(msg.sender, loss)
 
+        if strategy.doHealthCheck:
+            assert(pricePerShare * (MAX_BPS - strategy.healtyChanges) / MAX_BPS <= self._shareValue(10 ** self.decimals)) # dev: price decrease out of ranges
+        else:
+            self.strategies[msg.sender].doHealthCheck = True
```
fubuloubu, line 1700: Add a comment that we reset this boolean for next harvest

contracts/Vault.vy
steffenix: 👌thanks for the clarification!

contracts/Vault.vy
```diff
@@ -1294,6 +1298,18 @@ def updateStrategyPerformanceFee(
...
+def setStrategycheckLoss(strategy: address, enabled: bool):
+    assert msg.sender == self.governance 
+    assert self.strategies[strategy].activation > 0
+    self.strategies[strategy].checkLoss = enabled 
+
+@external
+def setStrategylossRatioLimit(strategy: address, percent: uint256):
+    assert msg.sender == self.governance 
+    assert self.strategies[strategy].activation > 0
+    self.strategies[strategy].lossRatioLimit = percent
```
fubuloubu, line 1311: add `assert percent <= MAX_BPS` here too

contracts/Vault.vy
```diff
@@ -1672,8 +1690,14 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
 
     # We have a loss to report, do it before the rest of the calculations
     if loss > 0:
+        pricePerShare: uint256 = self._shareValue(10 ** self.decimals)
```
fubuloubu, line 1693: Could probably more succinctly write this as the following:
```suggestion
        if self.strategies[msg.sender].checkLoss:
            assert loss <= self.strategies[msg.sender].lossRatioLimit * strategy_totalDebt / MAX_BPS
        else:
            self.strategies[msg.sender].checkLoss = True
```

fubuloubu (commented):
maybe measure overall loss instead of back-calculating share price change. it's a little clearer to work with

contracts/Vault.vy
fubuloubu: It's either normalizing the loss to a ratio and then to a price, or calculating what the end effect would be and backing out from there.

Either way, it's saying "revert if >3% impact."

contracts/Vault.vy
```diff
@@ -1294,6 +1299,19 @@ def updateStrategyPerformanceFee(
     log StrategyUpdatePerformanceFee(strategy, performanceFee)
 
 
+@external
+def setStrategycheckLoss(strategy: address, enabled: bool):
+    assert msg.sender == self.governance
```
dougstorm, line 1304: can be management or governance i would think

contracts/Vault.vy
```diff
@@ -1294,6 +1299,19 @@ def updateStrategyPerformanceFee(
...
 
+@external
+def setStrategycheckLoss(strategy: address, enabled: bool):
+    assert msg.sender == self.governance 
+    assert self.strategies[strategy].activation > 0
+    self.strategies[strategy].checkLoss = enabled 
+
+@external
+def setStrategylossRatioLimit(strategy: address, percent: uint256):
+    assert msg.sender == self.governance
```
dougstorm, line 1310: same, probably management or governance is good to have access to this

contracts/BaseStrategy.sol
```diff
@@ -16,6 +16,8 @@ struct StrategyParams {
     uint256 totalDebt;
     uint256 totalGain;
     uint256 totalLoss;
+    uint256 lossRatioLimit;
```
dougstorm, line 19: im torn about this since this breaks compatibility at the ABI and vault/ strat level, but not sure how to make it work other than adding a different method or struct for this or doing the change strictly at the BaseStrategy level only

contracts/Vault.vy
```diff
@@ -94,7 +94,8 @@ struct StrategyParams:
     totalDebt: uint256  # Total outstanding debt that Strategy has
     totalGain: uint256  # Total returns that Strategy has realized for Vault
     totalLoss: uint256  # Total losses that Strategy has realized for Vault
-
+    lossRatioLimit: uint256 # Allowed Percentage of price per share negative changes
```
dougstorm, line 97: is not just losses, can be a swing in profits thats too big or unexpected

contracts/Vault.vy
```diff
@@ -1148,6 +1149,7 @@ def addStrategy(
     minDebtPerHarvest: uint256,
     maxDebtPerHarvest: uint256,
     performanceFee: uint256,
+    lossRatioLimit: uint256 = 300 # 3%
```
dougstorm, line 1152: i think the main thing is a deviation both -/+ on the PPS pass a given threshold, there may be need for settings for both loss and profit.

contracts/BaseStrategy.sol
dougstorm: WDYT about adding healthcheck params to a different struct separated from the strategy params?

contracts/Vault.vy
```diff
@@ -1672,8 +1692,14 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
 
     # We have a loss to report, do it before the rest of the calculations
     if loss > 0:
+        pricePerShare: uint256 = self._shareValue(10 ** self.decimals)
         self._reportLoss(msg.sender, loss)
 
+        if self.strategies[msg.sender].checkLoss:
```
dougstorm, line 1698: I was wondering if a more full fledge healthcheck contract is needed to be deployed here to do this:

```
 if HealthCheck(healtcheckAddr).checkLoss(msg.sender):
      assert (HealthCheck(healtcheckAddr).healthCheck(msg.sender)) // sends vault addr + strat addr for global or special check logic

``` 



And we can turn the strategies on and off on that one, with no changes to struct on BaseStrategy, that way that Settings contract can have the healthcheck logic

dougstorm:
I feel this is good first draft, but maybe we should discuss having a separate contract do this for the vault and abstract away the health check since the vault is not easy migrated, this is not such a big issue if health check lives in strats since those are easily replaced.

contracts/Vault.vy
fubuloubu: I feel like this, coupled with your other comment https://github.com/yearn/yearn-vaults/pull/374#discussion_r639246982, points towards this being better implemented in `BaseStrategy`

fubuloubu: Also this comment https://github.com/yearn/yearn-vaults/pull/374#discussion_r639256575

instead of it living in another contract, just define it as invariants in `BaseStrategy`

contracts/BaseStrategy.sol
steffenix: Something that I have noticed, adding changes at the end of a structure won't break the existing code as long as these changes arent't needed of course.

I have made a PR just to run the tests without the changes to BaseStrategy to demonstrate that.

https://github.com/yearn/yearn-vaults/actions/runs/877678032

steffenix:
> I feel this is good first draft, but maybe we should discuss having a separate contract do this for the vault and abstract away the health check since the vault is not easy to migrate, this is not such a big issue if health check lives in strats since those are easily replaced.

I think the checks should be both in the strategy and the vault.

I see the check-in strategy as a check on as a view that is used with the `harvestTrigger` and prevents keeper from calling harvest. The vault is the final line of defense because it's the part that holds the final state. If we trust strategies with making sure it's not going to make big changes to the price per share, the issue remains the same, the code that checks if the strategy is healthy could be faulty.

contracts/Vault.vy
steffenix: Understood, I will make the changes to check price changes in both directions.

contracts/BaseStrategy.sol
fubuloubu: Yeah, actually this is true becuase of how ABI decoding works. It expects a particular size not a particular shape for a struct, and decodes it as basically a tuple.

contracts/Vault.vy
fubuloubu: Maybe change this to `changeLimitRatio`

fubuloubu:
> > I feel this is good first draft, but maybe we should discuss having a separate contract do this for the vault and abstract away the health check since the vault is not easy to migrate, this is not such a big issue if health check lives in strats since those are easily replaced.
> 
> I think the checks should be both in the strategy and the vault.
> 
> On the strategy, I see health check as a view function that is used with the `harvestTrigger` and prevents keeper from calling harvest.
> 
> The vault is the final line of defense because it's the part that holds the final state/price. If we trust strategies with making sure it's not going to make big changes and do not change the vault the issue remains the same if the code that checks the strategy health is faulty, we end up with a faulty price per share. The health check on the contract should always be the same regardless of the strategy it should make sure the change to the price is reasonable.

This is very true, however if we define this code in `BaseStrategy` as part of the default operation, it is nearly as good as it existing in `Vault`. We can also expand how it is used quite easily, with the strategy creator adding their own sanity checks that we can check for, while keeping the default excessive loss/gain check. I do get your points about the Vault being the last line of defense to a price increase/decrease from wrecking accounting, I just think it's really overkill to make this change here especially when other sanity checks will need to take place on a case-by-case basis anyways.

contracts/Vault.vy
fubuloubu: yeah, kind of nitpicky but it's like "change limit" that's a "ratio"

steffenix:
> This is very true, however, if we define this code in BaseStrategy as part of the default operation, it is nearly as good as it existing in Vault.

Will add to `harvest`

contracts/BaseStrategy.sol
```diff
@@ -869,9 +487,10 @@ abstract contract BaseStrategyInitializable is BaseStrategy {
         address _vault,
         address _strategist,
         address _rewards,
-        address _keeper
+        address _keeper,
+        uint256 _changeLimitRatio
```
dougstorm, line 491: lets add a sensible default and remove it from initialize, we can add a setter to update later

contracts/BaseStrategy.sol
```diff
@@ -885,19 +504,17 @@ abstract contract BaseStrategyInitializable is BaseStrategy {
...
-            // EIP-1167 bytecode
             let clone_code := mload(0x40)
             mstore(clone_code, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
             mstore(add(clone_code, 0x14), addressBytes)
             mstore(add(clone_code, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
             newStrategy := create(0, clone_code, 0x37)
         }
 
-        BaseStrategyInitializable(newStrategy).initialize(_vault, _strategist, _rewards, _keeper);
+        BaseStrategyInitializable(newStrategy).initialize(_vault, _strategist, _rewards, _keeper, changeLimitRatio);
```
dougstorm, line 517: same as above

contracts/BaseStrategy.sol
```diff
@@ -306,85 +266,55 @@ abstract contract BaseStrategy {
...
-     * @param _keeper The adddress of the _keeper. _keeper
-     * can harvest and tend a strategy.
-     */
     function _initialize(
         address _vault,
         address _strategist,
         address _rewards,
-        address _keeper
+        address _keeper,
+        uint256 _changeLimitRatio
```
dougstorm, line 277: prefer to add a sensible default so we don't break existing strat code and is easy to just extend from this one on existing strat code

contracts/BaseStrategy.sol
```diff
@@ -181,43 +179,16 @@ abstract contract BaseStrategy {
     using SafeMath for uint256;
     using SafeERC20 for IERC20;
     string public metadataURI;
+    uint256 constant MAX_BPS = 10_000;
+    uint256 public changeLimitRatio = 300;
```
dougstorm, line 183: we need a setter for onlyManagement = (governance or vault.management) modifier to update this in case of manual intervention

contracts/BaseStrategy.sol
```diff
@@ -306,85 +266,55 @@ abstract contract BaseStrategy {
...
 
-    /**
-     * @notice
-     *  Used to change `rewards`. EOA or smart contract which has the permission
-     *  to pull rewards from the vault.
-     *
-     *  This may only be called by the strategist.
-     * @param _rewards The address to use for pulling rewards.
-     */
+    function setStrategyEnforeChangeLimit(bool _enforeChangeLimit) external onlyAuthorized {
```
dougstorm, line 309: Can we create a new modifier that contains these roles governance or vault.management? since these two are msigs i believe this exceptional case should be hands on deck to every one of accepting a big price per share change.

Maybe add modifier to be named `onlyGovOrManagement`

contracts/BaseStrategy.sol
```diff
@@ -306,85 +266,55 @@ abstract contract BaseStrategy {
...
+        enforeChangeLimit = _enforeChangeLimit;
+    }
+
+    function setStrategychangeLimitRatio(uint256 _changeLimitRatio) external onlyAuthorized {
```
dougstorm, line 313: same modifier here

contracts/BaseStrategy.sol
```diff
@@ -734,118 +421,51 @@ abstract contract BaseStrategy {
...
-        // Check if free returns are left, and re-invest them
+        if (enforeChangeLimit) {
+            uint256 newPricePerShare = vault.pricePerShare();
+            require(((pricePerShare * (MAX_BPS - changeLimitRatio)) / MAX_BPS) <= newPricePerShare, "price decrease out of ranges");
+            require(((pricePerShare * (MAX_BPS + changeLimitRatio)) / MAX_BPS) >= newPricePerShare, "price increase out of ranges");
+        } else {
+            enforeChangeLimit = true;
+        }
+
         adjustPosition(debtOutstanding);
```
dougstorm, line 439: adjustPosition maybe should go above too

contracts/BaseStrategy.sol
```diff
@@ -566,166 +369,50 @@ abstract contract BaseStrategy {
...
-     * @dev
-     *  When `harvest()` is called, the Strategy reports to the Vault (via
-     *  `vault.report()`), so in some cases `harvest()` must be called in order
-     *  to take in profits, to borrow newly available funds from the Vault, or
-     *  otherwise adjust its position. In other cases `harvest()` must be
-     *  called to report to the Vault on the Strategy's position, especially if
-     *  any losses have occurred.
-     */
     function harvest() external onlyKeepers {
         uint256 profit = 0;
```
dougstorm, line 411: can we have an override method called, default does nothing but can be extended by strategist

`savePreHarvestState()`

and on line after 440 before ending the harvest call an override method called

`internalHealthCheck()` and default returns true and can be extended so we can use
`require(internalHealthCheck());` after PPS check

dougstorm: my only concernt above is it adds size to contract which some strats are already at max, the other way i can think is if we can return a dynamic uint256 array with 10 params in memory like

`uint256[] savedState = getPreHarvestState();`
// returns empty array by default up to strategist to pack in there what values they want to check
and give meaning to each index
...
`require(internalHealthCheck(savedState));`

Not sure if that makes sense

contracts/BaseStrategy.sol
```diff
@@ -734,118 +421,51 @@ abstract contract BaseStrategy {
...
-        // Allow Vault to take up to the "harvested" balance of this contract,
-        // which is the amount it has earned since the last time it reported to
-        // the Vault.
+        uint256 pricePerShare = vault.pricePerShare();
+
         debtOutstanding = vault.report(profit, loss, debtPayment);
 
-        // Check if free returns are left, and re-invest them
+        if (enforeChangeLimit) {
+            uint256 newPricePerShare = vault.pricePerShare();
```
dougstorm, line 432: i think you need to do a check directly on internals and locked profit because sandwich protection wont give you the real PPS value here

contracts/BaseStrategy.sol
```diff
@@ -734,118 +421,51 @@ abstract contract BaseStrategy {
...
             debtPayment = debtOutstanding.sub(loss);
         } else {
-            // Free up returns for Vault to pull
             (profit, loss, debtPayment) = prepareReturn(debtOutstanding);
         }
 
-        // Allow Vault to take up to the "harvested" balance of this contract,
-        // which is the amount it has earned since the last time it reported to
-        // the Vault.
+        uint256 pricePerShare = vault.pricePerShare();
```
Grandthrax, line 427: I think it would be better to use profit and loss instead of pricePerShare because pricePerShare won't change immediately due to sandwich protection.

totalDebt = vaul.totalDebt()
require(profit * 1e18 / totalDebt <= totalDebt  * (MAX_BPS - changeLimitRatio)) / MAX_BPS))
require(loss * 1e18 / totalDebt <= totalDebt  * (MAX_BPS - changeLimitRatio)) / MAX_BPS))

something like this. just make the threshold very small. like if we see a 0.1% increase or decrease there is issue

dougstorm: if @Grandthrax is a better measure of a big PPS change, by checking individual values (loss and profit) i think we go back to needing individual LimitRatios for each, since profitLimitRatio may be =0.05% but lossLimitRatio may be 0% for most if not all strats

fubuloubu: Agree with this, except just use each strategies' total debt instead, to keep it more normalized

```solidity
uint256 totalDebt = vaul.strategies(address(self)).totalDebt;
require(profit <= totalDebt * changeLimitRatio / MAX_BPS)
require(loss <= totalDebt * changeLimitRatio / MAX_BPS)
```

dougstorm:
Something that i just thought of, we need a way to port these changes of price checks to 0.3.0 and 0.3.2<->0.3.6 vaults @fubuloubu @steffenix.

Since 0.4.2 BaseStrategy is not compatible with those vault versions that are live and this is where this change will help the most on these live versions

steffenix:
closing in favor of https://github.com/yearn/yearn-vaults/pull/389

PR #375, opened 2021-05-25
feat: test failure

poolpitako:
Plz run `brownie test tests/functional/strategy/test_reduce_debt_ratio.py -sI`
I am getting a weird revert on `harvest()` but you can copy and paste and run in the cmd line.

You will see that after running the test `debtOutstanding` doesn't increase.


steffenix:
Looking at it

fubuloubu:
fix to be upstreamed in #381

Issue #376, opened 2021-05-26
DebtRatio problem


zgfzgf:
https://github.com/yearn/yearn-vaults/blob/ec7d2e314b438c1e87f1043e242dd562db998f4c/contracts/Vault.vy#L1441-L1445
https://github.com/yearn/yearn-vaults/blob/ec7d2e314b438c1e87f1043e242dd562db998f4c/contracts/Vault.vy#L1478
above. compute strategy_debtLimit  is different
https://github.com/yearn/yearn-vaults/blob/ec7d2e314b438c1e87f1043e242dd562db998f4c/contracts/Vault.vy#L1476
should add  var debtLimit
vault_debtLimit: uint256 =  self.debtLimit * vault_totalAssets / MAX_BPS 
self.debtLimit <= MAX_BPS

**No more judgment  self.debtRatio<=MAX_BPS**

compute strategy_debtLimit
 self.strategies[strategy].debtRatio   * self.totalDebt    / self.debtRatio  
add strategy
self.debtRatio +=  self.strategies[strategy].debtRatio

zgfzgf:
now use var self.debtRatio controller vault_debtLimit. It's hard to control and understand.

steffenix:
Yes, there is inconsistency plus we also have issues here
https://github.com/yearn/yearn-vaults/pull/375

steffenix:
What you are suggesting is the code we use to have prior to the changes.
https://github.com/yearn/yearn-vaults/commit/a5c51b29c80e19b588f8fbe35daf81e987368df3#diff-b583ba5c4d7b67e56b9084eff9fe883e12a2cdd94a2ea25029322c6f3568d807L1140

zgfzgf:
I think. add var debtLimit controller available amout.  vault 1 code:
// Custom logic in here for how much the vault allows to be borrowed
    // Sets minimum required on-hand to keep small withdrawals cheap
    function available() public view returns (uint) {
        return token.balanceOf(address(this)).mul(min).div(max);
    }
https://github.com/yearn/yearn-vaults/commit/a5c51b29c80e19b588f8fbe35daf81e987368df3#diff-b583ba5c4d7b67e56b9084eff9fe883e12a2cdd94a2ea25029322c6f3568d807L1140
there is no such function here.
add  var debtLimit.
We don't need that judgment
 self.debtRatio<=MAX_BPS
I think it's better and easier to understand. @steffenix @fubuloubu

steffenix:
[...]

PR #378, opened 2021-05-27
fix: delete assert self.debtRatio <= MAX_BPS

zgfzgf:
update self debt Ratio usage
This is a better way to control assets.
Solve a small part of this issue #376 

origin:
vault_debtLimit: uint256 = self.debtRatio * vault_totalAssets / MAX_BPS 
If this control is needed. we should add other var control it.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1181,7 +1181,6 @@ def addStrategy(
     assert self.token.address == Strategy(strategy).want()
 
     # Check strategy parameters
-    assert self.debtRatio + debtRatio <= MAX_BPS
     assert minDebtPerHarvest <= maxDebtPerHarvest
     assert performanceFee <= MAX_BPS / 2 
 
@@ -1225,7 +1224,6 @@ def updateStrategyDebtRatio(
     self.debtRatio -= self.strategies[strategy].debtRatio
     self.strategies[strategy].debtRatio = debtRatio
     self.debtRatio += debtRatio
-    assert self.debtRatio <= MAX_BPS
     log StrategyUpdateDebtRatio(strategy, debtRatio)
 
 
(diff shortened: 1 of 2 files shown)


steffenix:
We need to find a solution to this one https://github.com/yearn/yearn-vaults/pull/375 before going further on changing code related to debtRatio.

contracts/Vault.vy
```diff
@@ -1181,7 +1181,6 @@ def addStrategy(
     assert self.token.address == Strategy(strategy).want()
 
     # Check strategy parameters
-    assert self.debtRatio + debtRatio <= MAX_BPS
```
fubuloubu, line 1184: Not sure why you'd want to delete this, as it should ensure that `self.debtRatio <= 100%`, which is a necessary invariant for this code to work correctly

contracts/Vault.vy
```diff
@@ -1225,7 +1224,6 @@ def updateStrategyDebtRatio(
     self.debtRatio -= self.strategies[strategy].debtRatio
     self.strategies[strategy].debtRatio = debtRatio
     self.debtRatio += debtRatio
-    assert self.debtRatio <= MAX_BPS
```
fubuloubu, line 1228: Similar to above

contracts/Vault.vy
```diff
@@ -1472,10 +1470,12 @@ def _creditAvailable(strategy: address) -> uint256:
     # See note on `creditAvailable()`.
     if self.emergencyShutdown:
         return 0
+    if self.debtRatio == 0:
```
fubuloubu, line 1473: Could probably do this
```suggestion
    if self.debtRatio == 0 or self.strategies[strategy].debtRatio == 0:
```

But add caching for those two values e.g. `strategy_debtRatio` and `debtRatio` memory variables

contracts/Vault.vy
zgfzgf: update code. It is OK
https://github.com/yearn/yearn-vaults/blob/f5a576b8c9ae07130df28abedadcad90e49f93f0/contracts/Vault.vy#L1478-L1480
vault_debtLimit: uint256 =  vault_totalAssets
    vault_totalDebt: uint256 = self.totalDebt
    strategy_debtLimit: uint256 = self.strategies[strategy].debtRatio * vault_totalAssets / self.debtRatio

contracts/Vault.vy
zgfzgf: yes

zgfzgf:
core: self.debtRatio can't control asset. 
https://github.com/yearn/yearn-vaults/blob/f5a576b8c9ae07130df28abedadcad90e49f93f0/contracts/Vault.vy#L1478

PR #381, opened 2021-05-27
fix: revert changes on debt outstanding

steffenix:
fixes: #376


fubuloubu:
@poolpitako We discussed this issue and realized that the previous update to `_debtOutstanding` was incorrectly done, because that value should reflect the "expected value" of what the debt outstanding will be when `report` is called, and not the "current value" based on what the debt is (which won't update until next reporting cycle)

We had done this thinking there needed to be symmetry in how the calculations were performed. However, on further inspection we realized that the context of these two calculations were very different, and the approximation in the debt ratio adjustment in `-reportLoss` was okay to keep as we changed it in #337.

fubuloubu (approved):

tests/functional/strategy/test_misc.py
```diff
@@ -186,3 +186,12 @@ def test_set_metadataURI(gov, strategy, strategist, rando):
     assert strategy.metadataURI() == "ipfs://test3"
     with brownie.reverts("!authorized"):
         strategy.setMetadataURI("ipfs://fake", {"from": rando})
+
```
fubuloubu, line 189:
```suggestion


```

fubuloubu:
ignore commitlint failure, merge once CI passes

PR #383, opened 2021-05-28
perf: save gas in func withdraw

zgfzgf:
like #353


zgfzgf:
I suggest delete
https://github.com/yearn/yearn-vaults/blob/78c6b0d5afb4448fd40341decdd8b753a9846e90/contracts/Vault.vy#L1034-L1035

contracts/Vault.vy
```diff
@@ -1100,9 +1100,9 @@ def withdraw(
             #       including the losses that were incurred above during withdrawals
             shares = self._sharesForAmount(value + totalLoss)
 
-    # NOTE: This loss protection is put in place to revert if losses from
-    #       withdrawing are more than what is considered acceptable.
-    assert totalLoss <= maxLoss * (value + totalLoss) / MAX_BPS 
+        # NOTE: This loss protection is put in place to revert if losses from
+        #       withdrawing are more than what is considered acceptable.
+        assert totalLoss * MAX_BPS <= maxLoss * (value + totalLoss)
```
steffenix, line 1105: Why are these changes saving gas?

zgfzgf: see #353
* good /

steffenix: I don't see any benefits going from division to multiplication.

What you are pointing to in `2` is the change that moves the calculation within the if condition but I don't see a change of calculation.

On that PR only moving`totalLoss: uint256 = 0` make sense to me.

zgfzgf: update code   
https://github.com/yearn/yearn-vaults/blob/78c6b0d5afb4448fd40341decdd8b753a9846e90/contracts/Vault.vy#L1051
 should be below it

like #353

fubuloubu: I see what he's saying, but it's more clear if we keep `MAX_BPS` where it is, because it's adjusting the `maxLoss` parameter (which is really a ratio) to a value. From an auditing perspective, I prefer the clarity.

Revert this one reordering and this optimization looks good

fubuloubu (approved):

zgfzgf:
like #353    if value <= self.token.balanceOf(self): 
No need to:  assert totalLoss <= maxLoss * (value + totalLoss) / MAX_BPS

fubuloubu:
> like #353 if value <= self.token.balanceOf(self):
> No need to: assert totalLoss <= maxLoss * (value + totalLoss) / MAX_BPS

We're just saying we like `assert totalLoss <= maxLoss * (value + totalLoss) / MAX_BPS` because the terms make better sense together. The RHS and LHS both have units of "value" instead of "value * ratio"

zgfzgf:
Ok, assert totalLoss * MAX_BPS <= maxLoss * (value + totalLoss) 
rollback:
ssert totalLoss <= maxLoss * (value + totalLoss) / MAX_BPS

fubuloubu (approved):

PR #387, opened 2021-05-28
fix: keep initialize compatible with registry

steffenix:
reverts some of the changes made in #271


contracts/Vault.vy
```diff
@@ -235,10 +235,10 @@ def initialize(
     token: address,
     governance: address,
     rewards: address,
-    management: address,
     nameOverride: String[64],
     symbolOverride: String[32],
     guardian: address = msg.sender,
+    management: address = ZERO_ADDRESS,
```
fubuloubu, line 241: Can do `msg.sender` like `guardian`. There's no harm with it being set to the `Registry`, just requires more updates by governance.
```suggestion
    management: address = msg.sender,
```

contracts/Vault.vy
```diff
@@ -281,7 +281,10 @@ def initialize(
 
     self.governance = governance
     log UpdateGovernance(governance)
-    self.management = management
+    if management == ZERO_ADDRESS:
+        self.management = governance
+    else:
+        self.management = management
```
fubuloubu, line 287: See above, change not really needed here.

fubuloubu (approved):

PR #389, opened 2021-05-31
feat: add health check on base strategy

steffenix:


contracts/Vault.vy
```diff
@@ -1675,7 +1675,6 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     # We have a loss to report, do it before the rest of the calculations
     if loss > 0:
         self._reportLoss(msg.sender, loss)
-
```
dougstorm, line 1678: leftover spaces?

contracts/Vault.vy
```diff
@@ -1748,7 +1747,6 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
         # Otherwise, just return what we have as debt outstanding
         return debt
 
-
 @external
```
dougstorm, line 1750: leftover spaces?

contracts/BaseStrategy.sol
```diff
@@ -391,172 +327,51 @@ abstract contract BaseStrategy {
...
-     *  in that it is managing an active position, or will manage a position in
-     *  the future. This should correlate to `harvest()` activity, so that Harvest
-     *  events can be tracked externally by indexing agents.
-     * @return True if the strategy is actively managing a position.
-     */
     function isActive() public view returns (bool) {
         return vault.strategies(address(this)).debtRatio > 0 || estimatedTotalAssets() > 0;
     }
 
-    /**
```
dougstorm, line 536: is the linter removing all these docs in the methods?

contracts/BaseStrategy.sol
```diff
@@ -391,172 +327,51 @@ abstract contract BaseStrategy {
...
-     * NOTE: `_debtPayment` should be less than or equal to `_debtOutstanding`.
-     *       It is okay for it to be less than `_debtOutstanding`, as that
-     *       should only used as a guide for how much is left to pay back.
-     *       Payments should be made to minimize loss from slippage, debt,
-     *       withdrawal fees, etc.
-     *
-     * See `vault.debtOutstanding()`.
-     */
+    function healtyHarvest() public virtual view returns (bool) {
+        return true;
```
dougstorm, line 372: can you add more description around this, is the expectation that extending contract should override this logic?

Should it have override explicitly?

contracts/BaseStrategy.sol
```diff
@@ -734,119 +435,47 @@ abstract contract BaseStrategy {
             }
             debtPayment = debtOutstanding.sub(loss);
         } else {
-            // Free up returns for Vault to pull
             (profit, loss, debtPayment) = prepareReturn(debtOutstanding);
         }
+        if (enforeChangeLimit) {
+            uint256 totalDebt = vault.strategies(address(this)).totalDebt;
+            require(profit <= (totalDebt * changeLimitRatio) / MAX_BPS);
+            require(loss <= (totalDebt * changeLimitRatio) / MAX_BPS);
```
dougstorm, line 443: not sure if we need different values for each ratio to check, loss ratio may be 0% while profit ratio can be 1%

contracts/BaseStrategy.sol
fubuloubu: yeah wtf is up with this?

contracts/BaseStrategy.sol
steffenix: I think we should but would like other strategists to give their opinions. 
@retro-wavey

contracts/BaseStrategy.sol
steffenix: I added some doc but everything got removed, I have to see what's going on with the linter.
We can have two options for this:

```
    function healtyHarvest() public virtual view returns (bool) {
        return true;
    }
```

OR
```
    function healtyHarvest() public virtual view returns (bool)
```

If we go for the second one without the implementation that will force every strategy to implement the `healtyHarvest` function.

@Grandthrax @retro-wavey

contracts/BaseStrategy.sol
steffenix: running `yarn lint` removes the doc.

orbxball:
I don't think putting health check into base strategy is a good idea. Many current strategies used up 99% of code size. It's really hard to avoid "exceeds EIP-170 limit of 24577" error now. If we put more things in base strategy, it makes the situation worse. Including the health check and then no space to write strategy logic is meaningless, it's putting the cart before the horse. We should offload as many unnecessary variables from base strategy as possible, as long as these variables could be done else where. Strategy only takes care of one task, making profit. It doesn't have any spare attention/ability to take care of other things, e.g. accounting check, and it shouldn't because of modularity. 

There are 2 options. 
- check should be vault's job
- check could be offloaded into a separate external contract

The first option was dropped [here](https://github.com/yearn/yearn-vaults/pull/374) due to no backward compatibility. But if we can make the interface between vault and strategy backward compatible, I think it's ok to do it on vault side. And it's more reasonable from work division points of view. Vault handles accounting, so it has to do check too. Not strategy's responsiblity.

The second option is also a good direction. Since it's a separate contract, it's more adaptable and scalable. Now we need to add "accounting check" into it. Maybe later we need another check, and we could add into the check contract too.

dougstorm:
> od idea. Many current strategies used up 99% of code size. It's really hard to avoid "exceeds EIP-170 limit of 24577" error now. If we put more things in base strategy, it makes the situation worse. Including the health check and then no space to write strategy logic is meaningless, it's putting the cart before the horse. We should offload as many unnecessary variables from base strategy as possible, as long as these variables could be done else where. Strategy only takes care of one task, making profit. It doesn't have any spare attention/ability to take care of other things, e.g. accounting check, and it shouldn't because of modularity.

It sounds like a secondary/periphery contract is needed for this solution to be viable on all deployed strats since some of them are near the byte code size limit. Maybe we can explore a global healthcheck contract that we can inject to BaseStrat and call (gov only setter)?

fubuloubu:
> > od idea. Many current strategies used up 99% of code size. It's really hard to avoid "exceeds EIP-170 limit of 24577" error now. If we put more things in base strategy, it makes the situation worse. Including the health check and then no space to write strategy logic is meaningless, it's putting the cart before the horse. We should offload as many unnecessary variables from base strategy as possible, as long as these variables could be done else where. Strategy only takes care of one task, making profit. It doesn't have any spare attention/ability to take care of other things, e.g. accounting check, and it shouldn't because of modularity.
> 
> It sounds like a secondary/periphery contract is needed for this solution to be viable on all deployed strats since some of them are near the byte code size limit. Maybe we can explore a global healthcheck contract that we can inject to BaseStrat and call (gov only setter)?
> 
> I brought up this idea on first PR, maybe is good to discuss again:
> 
> [#374 (comment)](https://github.com/yearn/yearn-vaults/pull/374#discussion_r639256575)

That solution would also not be backwards compatible. If this is an issue for people, I think putting it in the Vault makes more sense.

dougstorm:
> > > od idea. Many current strategies used up 99% of code size. It's really hard to avoid "exceeds EIP-170 limit of 24577" error now. If we put more things in base strategy, it makes the situation worse. Including the health check and then no space to write strategy logic is meaningless, it's putting the cart before the horse. We should offload as many unnecessary variables from base strategy as possible, as long as these variables could be done else where. Strategy only takes care of one task, making profit. It doesn't have any spare attention/ability to take care of other things, e.g. accounting check, and it shouldn't because of modularity.
> > 
> > 
> > It sounds like a secondary/periphery contract is needed for this solution to be viable on all deployed strats since some of them are near the byte code size limit. Maybe we can explore a global healthcheck contract that we can inject to BaseStrat and call (gov only setter)?
> > I brought up this idea on first PR, maybe is good to discuss again:
> > [#374 (comment)](https://github.com/yearn/yearn-vaults/pull/374#discussion_r639256575)
> 
> That solution would also not be backwards compatible. If this is an issue for people, I think putting it in the Vault makes more sense.
why?
Can't we not do in BaseStrat instead of:

```

....

if (enforeChangeLimit) {            
uint256 totalDebt = vault.strategies(address(this)).totalDebt;            
require(profit &lt;= (totalDebt * changeLimitRatio) / MAX_BPS);            
require(loss &lt;= (totalDebt * changeLimitRatio) / MAX_BPS);
....

```

```
BaseStrategy.setHealthCheck(addr)
....

if (HealthCheck(addr).hasCheck()) {            
  // move all state needed for healthcheck downstream to external contract
   HealthCheck(addr).doHealthCheck(profit, loss, ....); 
   
....

```

fubuloubu:
> Basically the healthcheck becomes an external call and keep it backward compatible still targeting the BaseStrategy, later we can reuse same check contract on vault itself. Potentially a common healthcheck contract can be used across all strategy to start.

Either way you're using extra codesize in `BaseStrategy`

dougstorm:
> > Basically the healthcheck becomes an external call and keep it backward compatible still targeting the BaseStrategy, later we can reuse same check contract on vault itself. Potentially a common healthcheck contract can be used across all strategy to start.
> 
> Either way you're using extra codesize in `BaseStrategy`

Right, but the least added size to allow for the solution. I think for those 99%  size strats using this plus some additional refactoring may be needed too.

orbxball:
> > Basically the healthcheck becomes an external call and keep it backward compatible still targeting the BaseStrategy, later we can reuse same check contract on vault itself. Potentially a common healthcheck contract can be used across all strategy to start.
> 
> Either way you're using extra codesize in `BaseStrategy`

could we deploy a separate "check contract", and then register strategy addresses to it? So then health status could be checked on that "check contract" by reading off-chain, and this also dont consume any space on strategy, and need no changes on code of vault & strategy.

fubuloubu:
> > > Basically the healthcheck becomes an external call and keep it backward compatible still targeting the BaseStrategy, later we can reuse same check contract on vault itself. Potentially a common healthcheck contract can be used across all strategy to start.
> > 
> > 
> > Either way you're using extra codesize in `BaseStrategy`
> 
> How about deploy a separate "check contract", and then register strategy address to it? So then health status could be checked on that "check contract" by reading either off-chain or on-chain, and this also doesnt consume any space on strategy, and need no changes on both vault code and strategy code.

The point of the feature is to revert when a health check is violated, so some code change is necessary in either `Vault` or `BaseStrategy`

dougstorm:
> Ill try to test both solutions in a strategy with 99% code size and come back to see what options is feasible on these two repos:
> https://github.com/orbxball/liquity-stability-pool
> https://github.com/jmonteer/yearnV2-aave-lender-borrower

Did the testing on both approaches i mentioned on a clone repo from aave-lender and here are the two PRs comparing contract size after fix:

Fix in BaseStrategy:
https://github.com/dougstorm/yearnV2-aave-lender-borrower/pull/1

Fix using external contract call in BaseStrategy
https://github.com/dougstorm/yearnV2-aave-lender-borrower/pull/2

TLDR: seems the second option allows us to make the fix without affecting the contract size in a meaningful way.
NOTE: the PRs for both are quick hacking, feel free to copy and add testing to my sample code since i didnt test it just checked compilation and contract size @steffenix @fubuloubu @orbxball

steffenix:
I still in favor of checks as part of the vault and not the strategy despite the lack of backward compatibility.

Here are my suggestions:

- version 0.3 adds checks to BaseStrategy as an external library and keeps compatibility.
- version 0.4 add checks to the vault
- For strategy that has reached the size limit, they should be prioritized to migrate the vault to 0.4 with the checks

Regarding contract size @orbxball we have prepared solidity 0.8 changes that get rid of SafeMath, later we could explorer a solution with a StrategyLibrary that should reduce the size of the contract.

@dougstorm @fubuloubu

jmonteer:
Chiming in to give my strategist POV. 

I agree pretty much with @orbxball 's opinion. I have hit the bytecode size limit and it is super painful and leads to inconvenient code (like merging setters together, reducing visibility of state variables, ...). I believe we should aim for a strategy to exclusively worry about making money. This means using external contracts to handle unrelated logic, in which I would include health checking but also swaps and other things that have to be done right and which does not make sense to reinvent each time... 

As a parentheses: this might be a good moment to try to rethink BaseStrategy and make it lighter weight. 

IMO, best option if we started from scratch would be for the vault to handle all accounting logic (thus, its sanity checking too). Taking into account that it would mean migrating vaults, which are the hardest to migrate, I would use the solution proposed by @dougstorm , which calls an external contract that handles healthcheck logic during harvest. This requires upgrade and re-deployment for all existing strategies, which is a pain. 
 
Another alternative is to just solve the long-term part of this, add the healthcheck (or an external call to a healthcheck contract) to the vault and solve it for future iterations. We avoid strategy upgrades and migration and, in the meanwhile, we rely on off-chain monitoring (which may result in some issues but will probably catch most of them) for those vaults that have not been yet migrated. That way we apply what I believe we agree is the best solution (vault should handle this), and mitigate as much as possible new cases for old vaults.

jmonteer:
> I still in favor of checks as part of the vault and not the strategy despite the lack of backward compatibility.
> 
> Here are my suggestions:
> 
> * version 0.3 adds checks to BaseStrategy as an external library and keeps compatibility.
> * version 0.4 add checks to the vault
> * For strategy that has reached the size limit, they should be prioritized to migrate the vault to 0.4 with the checks
> 
> Regarding contract size @orbxball we have prepared solidity 0.8 changes that get rid of SafeMath, later we could explorer a solution with a StrategyLibrary that should reduce the size of the contract.
> 
> @dougstorm @fubuloubu

Agree with this. 

only difference is that I would also consider throwing away the 0.3 on-chain part, I see this as a good compromise between efforts and risks.

dougstorm:
> I still in favor of checks as part of the vault and not the strategy despite the lack of backward compatibility.
> 
> Here are my suggestions:
> 
> * version 0.3 adds checks to BaseStrategy as an external library and keeps compatibility.
> * version 0.4 add checks to the vault
> * For strategy that has reached the size limit, they should be prioritized to migrate the vault to 0.4 with the checks
> 
> Regarding contract size @orbxball we have prepared solidity 0.8 changes that get rid of SafeMath, later we could explorer a solution with a StrategyLibrary that should reduce the size of the contract.
> 
> @dougstorm @fubuloubu

Agree, only thing is that 0.3 solution is an external contract and not a library but it works the same.

Strategies that have 99% can have workarounds (refactor public vars, etc), and in any case i tested the fix with external contracts and it doesnt affect the size of the strategy at all, see my post above.

retro-wavey:
a thought: 
do we have an opportunity here to make further savings on compile size? We could off-load, for example, the default harvestTrigger() implementation to this separate contract as well.

steffenix:
> a thought:
> do we have an opportunity here to make further savings on compile size? For example, we could off-load the default `harvestTrigger()` implementation to this separate contract as well.

Yes, we will consider that on top of 0.4 or 1.0.

contracts/BaseStrategy.sol
Grandthrax: yes please. default should be 0 for loss in my opinion. and strategist can raise it if it is an unusual strategy that has regular losses

steffenix:
Heath check has been added to versions from 0.3.0 to 0.3.5.
On version 0.4 the checks will be added to the vault.

PR #394, opened 2021-06-02
fix: flash deposit withdraw

zgfzgf:
#376 
@steffenix say:“
Take a flash loan
Make deposit
Call report
Make a withdrawal
Repay the flash loan
”


steffenix:
No fund locks please.

steffenix:
One more thing, this is still a invalid protection, one user can do any number of deposits just under the `minLockedAmount`.

zgfzgf:
> One more thing, this is still a invalid protection, one user can do any number of deposits just under the `minLockedAmount`.

Yes. Just be relatively safe. 
when minLockedAmoun = 0. all deposit locked.
It's just an implementation to solve the problem. @steffenix @fubuloubu

fubuloubu:
At first glance this doesn't appear very sybil resistant. You can have multiple flashloans for multiple deposits and bypass this check.

PR #395, opened 2021-06-03
fix: update pendingGovernance public

zgfzgf:
https://github.com/compound-finance/compound-protocol/blob/b9b14038612d846b83f8a009a82c38974ff2dcfe/contracts/Timelock.sol#L20
https://github.com/compound-finance/compound-protocol/blob/b9b14038612d846b83f8a009a82c38974ff2dcfe/contracts/Timelock.sol#L48

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -82,7 +82,7 @@ token: public(ERC20)
 governance: public(address)
 management: public(address)
 guardian: public(address)
-pendingGovernance: address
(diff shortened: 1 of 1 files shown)


zgfzgf:
The cause of the test error was not found

steffenix:
Sorry, I have no idea what you are talking about?

zgfzgf:
@steffenix now OK. The test is not stable

PR #399, opened 2021-06-05
fix: apply performance fee only to gain on undelegated assets

retro-wavey:
Right now governance takes an extra 10% performance fee from strategies that delegate. This fix uses same methodology used to compute management fee based on delegated assets, and applies it to performance fee to prevent a double-dip, making delegated strategies more viable.

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1602,7 +1602,7 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
             / SECS_PER_YEAR
         )
 
-         # NOTE: Unlikely to throw unless strategy reports >1e72 harvest profit
+        # NOTE: Unlikely to throw unless strategy reports >1e72 harvest profit
         performance_fee = (
             # NOTE: Fee is applied to gain in same ratio as we have for undelegated assets to total assets
             gain
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1585,16 +1585,32 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
     if gain == 0:
         # NOTE: The fees are not charged if there hasn't been any gains reported
         return 0
+    
+    undelegatedAssets: uint256 = self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets()
 
-    management_fee: uint256 = (
-        (
-            (self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets())
-            * duration 
-            * self.managementFee
+    performance_fee: uint256 = 0
+    management_fee: uint256 = 0
+
+    if undelegatedAssets > 0:
+        management_fee = (
+            (
+                undelegatedAssets
+                * duration 
+                * self.managementFee
+            )
+            / MAX_BPS
+            / SECS_PER_YEAR
         )
-        / MAX_BPS
-        / SECS_PER_YEAR
-    )
+
+         # NOTE: Unlikely to throw unless strategy reports >1e72 harvest profit
+        performance_fee = (
+            # NOTE: Fee is applied to gain in same ratio as we have for undelegated assets to total assets
+            gain
+            * self.performanceFee
+            * undelegatedAssets
+            / self.strategies[strategy].totalDebt
+            / MAX_BPS
+        )
 
     # NOTE: Applies if Strategy is not shutting down, or it is but all debt paid off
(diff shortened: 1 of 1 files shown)


steffenix:
I don't know if that's fair to reduce the performance_fee but leave the strategist_fee untouched.

retro-wavey:
> Is that's fair to reduce the performance_fee but leave the strategist_fee untouched?

It's a good question, and a hot topic right now. We still have some issues to work thru regarding where the strategist fee should be applied. It's a tricky area since often the delegating strategies are more complex to write than the ones which earn the yield. We want to make sure incentives are properly aligned.

But for this particular PR, I view this change as a "quick-win" to reduce performance fee double-dipping. The treasury address is the same recipient on all vaults, so it doesn't matter which vault applies the fee. Just as long 1 and only 1 vault does.

contracts/Vault.vy
```diff
@@ -1585,16 +1585,32 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
...
-            (self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets())
-            * duration 
-            * self.managementFee
+         # NOTE: Unlikely to throw unless strategy reports >1e72 harvest profit
+        performance_fee = (
+            # NOTE: Fee is applied to gain in same ratio as we have for undelegated assets to total assets
+            gain
+            * self.performanceFee
+            * undelegatedAssets
+            / self.strategies[strategy].totalDebt
```
fubuloubu, line 1611: Maybe cache this value too?

contracts/Vault.vy
```diff
@@ -1585,16 +1585,32 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
...
+            / SECS_PER_YEAR
+        )
 
-    management_fee: uint256 = (
-        (
-            (self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets())
-            * duration 
-            * self.managementFee
+         # NOTE: Unlikely to throw unless strategy reports >1e72 harvest profit
+        performance_fee = (
```
fubuloubu, line 1606: If governance gets 0 fee for fully delegated strategy, but strategist gets the full fee, it's kind of a weird situation.

It might be better to have more control over this value by moving the performance fee to governance to being independently set per-strategy

contracts/Vault.vy
```diff
@@ -1585,16 +1585,32 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
...
+            / MAX_BPS
+            / SECS_PER_YEAR
+        )
 
-    management_fee: uint256 = (
-        (
-            (self.strategies[strategy].totalDebt - Strategy(strategy).delegatedAssets())
-            * duration 
-            * self.managementFee
+         # NOTE: Unlikely to throw unless strategy reports >1e72 harvest profit
```
fubuloubu, line 1605:
```suggestion
        # NOTE: Unlikely to throw unless strategy reports >1e72 harvest profit
```

contracts/Vault.vy
retro-wavey: My preference would be that all fees (gov perf, strategist perf, and mgmt) get handled using the delegation logic. To me it is the more elegant solution: is automatic, ensures consistency, and requires less management overhead.

Only problem right now is that we don't have a great framework for how to handle the strategist fee on delegated strats.  

What do you think?

fubuloubu: We don't have *any* framework for handling these delegated fund strategies currently. We add the mgmt fee fix as a "user benefit" since we see the mgmt fee as being core profit for our Vaults. The performance fee sort of pays for overhead costs, loosely, so that's why we didn't make the change to the performance fee. The strategists also still put in effort to develop the strategy, which is why we didn't adjust that fee either.

retro-wavey: Sure. I agree. Fees overall need a fresh look. Either we stay true to 2/20 or we adjust to a new scheme that makes more sense. 

For now, with yields so low, i worry that fees in excess of 2/20 will result in extremely low (or zero) yield to user, harming our TVL. all Maker strats right now are at or near net 0 yield for user.

fubuloubu: Have we considered... disabling the Maker strats?

retro-wavey: as long as they're earning revenue for us and not net negative, we should keep them enabled.

PR #402, opened 2021-06-08
feat: add description field

retro-wavey:


steffenix:
Could you update files in https://github.com/yearn/yearn-devdocs/blob/master/docs/v2/

steffenix (approved):

Issue #152, opened 2021-01-14
Explore more efficient data structure for Wthdrawal Queue


fubuloubu:
From Audit Report:

The vault contract manages a withdraw queue. In some locations duplicate checks are being performed. However, in the function `Vault.setWithdrawalQueue(_newQueue)` there are protections against adding the address zero but there is no protection against adding the same address multiple times to the queue.

Additionally, the function `Vault._organizeWithdrawalQueue()` reorganizes the withdrawal queue such that all zero addresses are at the end of the array and none is in-between valid addresses. When adding a strategy to the queue by calling `Vault,addStrategyToQueue(_strategy, *params)` the loop ensures that the new strategy to be added is not in the queue already. We should evaluate to use the loop to put the new strategy into the correct free slot already. This would save the reorganization of the queue done in the next step with `Vautl._organizeWithdrawalQueue.()`

---

Explore a data structure that could enable constant-time membership checks as well as linear time insertion/removal/iteration.

NOTE: research adding this directly to Vyper as an additional data type available e.g. `OrderedSet`

steffenix:
The first part of this issue is resolved by https://github.com/yearn/yearn-vaults/pull/299

Niraj-Kamdar:
I have created an in-memory set data-structure with simple hash function which can be used to remove duplicates from an array. 
You can checkout the gist at: https://gist.github.com/Niraj-Kamdar/25d47ae6773ea702cc583a1403627cdd

I will profile it for gas cost and if it seems better than the current solution then we can replace it.

steffenix:
@Niraj-Kamdar in our case, this needs to be on our vault, it's written in Vyper.

Niraj-Kamdar:
I think I will be able to convert this into a Vyper code.

fubuloubu:
> I have created an in-memory set data-structure with simple hash function which can be used to remove duplicates from an array.
> You can checkout the gist at: https://gist.github.com/Niraj-Kamdar/25d47ae6773ea702cc583a1403627cdd
> 
> I will profile it for gas cost and if it seems better than the current solution then we can replace it.

Thanks for sharing! As @steffenix said, in order to work for us it would need to be implemented in Vyper. but could also explore the idea of set membership to maintain proper use of the queue.

Niraj-Kamdar:
I will try to convert it into vyper and make it work out for yearn vault

Niraj-Kamdar:
[...]

PR #404, opened 2021-06-09
feat: more efficient data structure for checking Withdrawal Queue 

Niraj-Kamdar:
related: #152

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -567,12 +567,11 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
 
         assert self.strategies[queue[i]].activation > 0
 
-        # hash will have hash of the address in uint256.
-        # `mask = SET_SIZE - 1` will be used to get key for the set for indexing
+        # NOTE: `key` is first `log_2(SET_SIZE)` bits of address (which is a hash)
         hash: uint256 = convert(queue[i], uint256)
-        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `log_2(SET_SIZE)` bits of address
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  
         # Most of the times following for loop only run once which is making it highly gas efficient 
-        # but in the worst case of key collision it will run linearly and find first empty slot.
+        # but in the worst case of key collision it will run linearly and find first empty slot in the set.
         for j in range(key, key + SET_SIZE):
             idx: uint256 = j % SET_SIZE  # traverse set for empty space in circular manner
             assert set[idx] != queue[i]  # dev: duplicate in set
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -192,7 +192,7 @@ event StrategyAddedToQueue:
 strategies: public(HashMap[address, StrategyParams])
 MAXIMUM_STRATEGIES: constant(uint256) = 20
 DEGRADATION_COEFFICIENT: constant(uint256) = 10 ** 18
-SET_SIZE: constant(uint256) = 2 * MAXIMUM_STRATEGIES
+SET_SIZE: constant(uint256) = 32 # 2 * MAXIMUM_STRATEGIES
 
 # Ordering that `withdraw` uses to determine which strategies to pull funds from
 # NOTE: Does *NOT* have to match the ordering of all the current strategies that
@@ -533,27 +533,6 @@ def getHash(val: address) -> uint256:
     return convert(val, uint256)
 
 
-# @internal
-# def checkAndInsertInSet(set: address[SET_SIZE], val: address) -> bool:
-#     hash: uint256 = self.getHash(val)
-#     mask: uint256 = SET_SIZE - 1
-#     key: uint256 = bitwise_and(hash, mask)
-#     assert key < SET_SIZE
-#     for i in range(SET_SIZE):
-#         if set[i] == val: 
-#             return True
-#         if set[i] == ZERO_ADDRESS:
-#             set[i] = val
-#             return False
-#     for i in range(0, key):
-#         if set[i] == val:
-#             return True
-#         if set[i] == ZERO_ADDRESS:
-#             set[i] = val
-#             return False
-#     assert False
-
-
 @external
 def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
     """
@@ -597,20 +576,22 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         mask: uint256 = SET_SIZE - 1
         key: uint256 = bitwise_and(hash, mask)
         assert key < SET_SIZE
-        for k in range(key, key + SET_SIZE):
-            if k >= SET_SIZE:
+        for m in range(key, key + SET_SIZE):
+            if m >= SET_SIZE:
+                for n in range(SET_SIZE):
+                    if n >= key:
+                        break
+                    if set[n] == queue[i]:
+                        assert False
+                    if set[n] == ZERO_ADDRESS:
+                        set[n] = queue[i]
+                        break
                 break
-            if set[k] == queue[i]: 
+            if set[m] == queue[i]: 
                 assert False
-            if set[k] == ZERO_ADDRESS:
-                set[k] = queue[i]
-        for k in range(SET_SIZE):
-            if k >= key:
+            if set[m] == ZERO_ADDRESS:
+                set[m] = queue[i]
                 break
-            if set[k] == queue[i]:
-                assert False
-            if set[k] == ZERO_ADDRESS:
-                set[k] = queue[i]
 
         self.withdrawalQueue[i] = queue[i]
 
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -573,21 +573,11 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `log_2(SET_SIZE)` bits of address
         # Most of the times following for loop only run once which is making it highly gas efficient 
         # but in the worst case of key collision it will run linearly and find first empty slot.
-        for m in range(key, key + SET_SIZE):
-            # We've exhausted the set but still need to place this item
-            if m >= SET_SIZE:
-                # So we're starting over from the beginning of the set to find space in the set
-                for n in range(SET_SIZE):
-                    # Note: This only means SET_SIZE is lower than MAXIMUM_STRATEGIES
-                    assert n < key
-                    assert set[n] != queue[i]
-                    if set[n] == ZERO_ADDRESS:
-                        set[n] = queue[i]
-                        break
-                break
-            assert set[m] != queue[i]
-            if set[m] == ZERO_ADDRESS:
-                set[m] = queue[i]
+        for j in range(key, key + SET_SIZE):
+            idx: uint256 = j % SET_SIZE  # traverse set for empty space in circular manner
+            assert set[idx] != queue[i]  # duplicate check
+            if set[idx] == ZERO_ADDRESS:
+                set[idx] = queue[i]
                 break
 
         self.withdrawalQueue[i] = queue[i]
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -574,6 +574,7 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         # Most of the times following for loop only run once which is making it highly gas efficient 
         # but in the worst case of key collision it will run linearly and find first empty slot.
         for m in range(key, key + SET_SIZE):
+            # We've exhausted the set but still need to place this item
             if m >= SET_SIZE:
                 for n in range(SET_SIZE):
                     # Note: This only means SET_SIZE is lower than MAXIMUM_STRATEGIES
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -575,7 +575,7 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         # but in the worst case of key collision it will run linearly and find first empty slot.
         for j in range(key, key + SET_SIZE):
             idx: uint256 = j % SET_SIZE  # traverse set for empty space in circular manner
-            assert set[idx] != queue[i]  # duplicate check
+            assert set[idx] != queue[i]  # dev: duplicate in set
             if set[idx] == ZERO_ADDRESS:
                 set[idx] = queue[i]
                 break
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -573,7 +573,8 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         # Most of the times following for loop only run once which is making it highly gas efficient 
         # but in the worst case of key collision it will run linearly and find first empty slot in the set.
         for j in range(key, key + SET_SIZE):
-            idx: uint256 = j % SET_SIZE  # traverse set for empty space in circular manner
+            # NOTE: we can always find space by treating set as circular (as long as `SET_SIZE >= MAXIMUM_STRATEGIES`)
+            idx: uint256 = j % SET_SIZE
             assert set[idx] != queue[i]  # dev: duplicate in set
             if set[idx] == ZERO_ADDRESS:
                 set[idx] = queue[i]
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -571,9 +571,9 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         key: uint256 = bitwise_and(convert(queue[i], uint256), SET_SIZE - 1)  
         # Most of the times following for loop only run once which is making it highly gas efficient 
         # but in the worst case of key collision it will run linearly and find first empty slot in the set.
-        for j in range(key, key + SET_SIZE):
+        for j in range(SET_SIZE):
             # NOTE: we can always find space by treating set as circular (as long as `SET_SIZE >= MAXIMUM_STRATEGIES`)
-            idx: uint256 = j % SET_SIZE
+            idx: uint256 = (key + j) % SET_SIZE
             assert set[idx] != queue[i]  # dev: duplicate in set
             if set[idx] == ZERO_ADDRESS:
                 set[idx] = queue[i]
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -568,8 +568,7 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         assert self.strategies[queue[i]].activation > 0
 
         # NOTE: `key` is first `log_2(SET_SIZE)` bits of address (which is a hash)
-        hash: uint256 = convert(queue[i], uint256)
-        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  
+        key: uint256 = bitwise_and(convert(queue[i], uint256), SET_SIZE - 1)  
         # Most of the times following for loop only run once which is making it highly gas efficient 
         # but in the worst case of key collision it will run linearly and find first empty slot in the set.
         for j in range(key, key + SET_SIZE):
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -192,8 +192,8 @@ event StrategyAddedToQueue:
 strategies: public(HashMap[address, StrategyParams])
 MAXIMUM_STRATEGIES: constant(uint256) = 20
 DEGRADATION_COEFFICIENT: constant(uint256) = 10 ** 18
-# SET_SIZE can be any number but having it in power of 2 will be more gas friendly 
-# and collision free.
+# SET_SIZE can be any number but having it in power of 2 will be more gas friendly and collision free. 
+# Note: Make sure SET_SIZE is greater than MAXIMUM_STRATEGIES 
 SET_SIZE: constant(uint256) = 32 
 
 # Ordering that `withdraw` uses to determine which strategies to pull funds from
@@ -568,26 +568,22 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
         assert self.strategies[queue[i]].activation > 0
 
         # hash will have hash of the address in uint256.
-        # mask will be used to get key for the set for indexing
+        # `mask = SET_SIZE - 1` will be used to get key for the set for indexing
         hash: uint256 = convert(queue[i], uint256)
-        mask: uint256 = SET_SIZE - 1
-        key: uint256 = bitwise_and(hash, mask)
-        assert key < SET_SIZE
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `SET_SIZE` bits of address
         # Most of the times following for loop only run once which is making it highly gas efficient 
         # but in the worst case of key collision it will run linearly and find first empty slot.
         for m in range(key, key + SET_SIZE):
             if m >= SET_SIZE:
                 for n in range(SET_SIZE):
-                    if n >= key:
-                        break
-                    if set[n] == queue[i]:
-                        assert False
+                    # Note: This only means SET_SIZE is lower than MAXIMUM_STRATEGIES
+                    assert n < key
+                    assert set[n] != queue[i]
                     if set[n] == ZERO_ADDRESS:
                         set[n] = queue[i]
                         break
                 break
-            if set[m] == queue[i]: 
-                assert False
+            assert set[m] != queue[i]
             if set[m] == ZERO_ADDRESS:
                 set[m] = queue[i]
                 break
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -192,6 +192,7 @@ event StrategyAddedToQueue:
 strategies: public(HashMap[address, StrategyParams])
 MAXIMUM_STRATEGIES: constant(uint256) = 20
 DEGRADATION_COEFFICIENT: constant(uint256) = 10 ** 18
+SET_SIZE: constant(uint256) = 2 * MAXIMUM_STRATEGIES
 
 # Ordering that `withdraw` uses to determine which strategies to pull funds from
 # NOTE: Does *NOT* have to match the ordering of all the current strategies that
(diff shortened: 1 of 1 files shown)


contracts/Vault.vy
```diff
@@ -553,36 +556,44 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+        assert key < SET_SIZE
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot.
+        for m in range(key, key + SET_SIZE):
+            if m >= SET_SIZE:
+                for n in range(SET_SIZE):
+                    if n >= key:
+                        break
+                    if set[n] == queue[i]:
+                        assert False
```
Niraj-Kamdar, line 584: duplicate found

fubuloubu:
```suggestion
                        raise
```

fubuloubu:
> ~~fixes~~ related: #152

I changed this to only being related, because we wanted to see if a number of other things might be possible e.g. unlimited size queue, less gas costs, etc.

Niraj-Kamdar:
Can you tell me how to fix the commit lint issues?

contracts/Vault.vy
```diff
@@ -553,36 +556,45 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
 
-        existsInOldQueue: bool = False
-        for j in range(MAXIMUM_STRATEGIES):
-            if queue[j] == ZERO_ADDRESS:
-                existsInOldQueue = True
+        # hash will have hash of the address in uint256.
+        # mask will be used to get key for the set for indexing
+        hash: uint256 = convert(queue[i], uint256)
+        mask: uint256 = SET_SIZE - 1
+        key: uint256 = bitwise_and(hash, mask)
```
fubuloubu, line 574:
```suggestion
        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `SET_SIZE` bits of address
```

contracts/Vault.vy
```diff
@@ -553,36 +556,45 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+        assert key < SET_SIZE
```
fubuloubu, line 575: By definition this is correct
```suggestion
```

contracts/Vault.vy
```diff
@@ -553,36 +556,45 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot.
+        for m in range(key, key + SET_SIZE):
+            if m >= SET_SIZE:
+                for n in range(SET_SIZE):
+                    if n >= key:
+                        # Note: This only means SET_SIZE is lower than MAXIMUM_STRATEGIES
+                        raise
```
fubuloubu, line 583: Prefer this
```suggestion
                    # Note: This only means SET_SIZE is lower than MAXIMUM_STRATEGIES
                    assert n < key
```

contracts/Vault.vy
```diff
@@ -553,36 +556,45 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+                    if set[n] == queue[i]:
+                        raise
```
fubuloubu, line 585:
```suggestion
                    assert set[n] != queue[i]
```

contracts/Vault.vy
```diff
@@ -553,36 +556,45 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+                    if set[n] == ZERO_ADDRESS:
+                        set[n] = queue[i]
+                        break
+                break
+            if set[m] == queue[i]: 
+                raise
```
fubuloubu, line 591:
```suggestion
            assert set[m] == queue[i]
```

fubuloubu (commented):
What happens when there is a hash collision in the set because only 5 bits of entropy (`2**5 = 32`) are used?

Niraj-Kamdar:
> What happens when there is a hash collision in the set because only 5 bits of entropy (`2**5 = 32`) are used?

Since MAXIMUM_STRATEGIES is 20,  SET_SIZE used is 32 (nearest bigger power of 2)
Whenever hash collision happen. We will find the next empty slot in the SET which will take linear time and since SET_SIZE will always be greater than MAXIMUM_STRATEGIES, it will be able to find an empty slot for sure. Increasing the SET_SIZE can increase the performance if there are many collision happening but probability of multiple collision reduces very quickly. So, I don't think it will be an issue. 

If we want to be extra cautious then we can hash all the possible addresses and find out exactly how many collisions can happen.

fubuloubu:
> > What happens when there is a hash collision in the set because only 5 bits of entropy (`2**5 = 32`) are used?
> 
> Since MAXIMUM_STRATEGIES is 20, SET_SIZE used is 32 (nearest bigger power of 2)
> Whenever hash collision happen. We will find the next empty slot in the SET which will take linear time and since SET_SIZE will always be greater than MAXIMUM_STRATEGIES, it will be able to find an empty slot for sure. Increasing the SET_SIZE can increase the performance if there are many collision happening but probability of multiple collision reduces very quickly. So, I don't think it will be an issue.
> 
> If we want to be extra cautious then we can hash all the possible addresses and find out exactly how many collisions can happen.

Ah, I think I get it now. Basically, it just orders the set by the first 5 bits of the hash.

Honestly, the address is already a hash, could probably just take the first 5 bits of that

Niraj-Kamdar:
> Honestly, the address is already a hash, could probably just take the first 5 bits of that
Don't we need to convert it to uint?
If it would be possible to get the last 5 bits from the address directly. It will be super efficient.

fubuloubu:
> Can you tell me how to fix the commit lint issues?

`git rebase -i master` and then edit or squash commits as needed

fubuloubu:
> > Honestly, the address is already a hash, could probably just take the first 5 bits of that
> > Don't we need to convert it to uint?
> > If it would be possible to get the last 5 bits from the address directly. It will be super efficient.

Just bitshift by 155 bits (address is bytes20 aka 160 bits, but you only need 5 bits for the ordered set)

Niraj-Kamdar:
> Just bitshift by 155 bits (address is bytes20 aka 160 bits, but you only need 5 bits for the ordered set)
I will try this out and compare the gas profile.

fubuloubu:
> > Just bitshift by 155 bits (address is bytes20 aka 160 bits, but you only need 5 bits for the ordered set)
> 
> I will try this out and compare the gas profile.

Oh wait, you are already doing this lol. I thought you added an extra hash in there, which is like 700 gas

Niraj-Kamdar:
> Oh wait, you are already doing this lol. I thought you added an extra hash in there, which is like 700 gas
Nope just type casting and taking bitwise_and

contracts/Vault.vy
```diff
@@ -553,36 +556,40 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+        # `mask = SET_SIZE - 1` will be used to get key for the set for indexing
+        hash: uint256 = convert(queue[i], uint256)
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `SET_SIZE` bits of address
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot.
+        for m in range(key, key + SET_SIZE):
+            if m >= SET_SIZE:
+                for n in range(SET_SIZE):
+                    # Note: This only means SET_SIZE is lower than MAXIMUM_STRATEGIES
+                    assert n < key
```
fubuloubu, line 580: Would this be appropriate, considering the above comment?
```suggestion
                    assert n <= MAXIMUM_STRATEGIES
```

Niraj-Kamdar: Nope it should be `<key` because `>= key` has been traversed already. Comment is for understanding the situation better.

contracts/Vault.vy
```diff
@@ -553,36 +556,40 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
-            if queue[j] == ZERO_ADDRESS:
-                existsInOldQueue = True
+        # hash will have hash of the address in uint256.
+        # `mask = SET_SIZE - 1` will be used to get key for the set for indexing
+        hash: uint256 = convert(queue[i], uint256)
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `SET_SIZE` bits of address
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot.
+        for m in range(key, key + SET_SIZE):
+            if m >= SET_SIZE:
```
fubuloubu, line 577: Add a note about why this works

contracts/Vault.vy
```diff
@@ -553,36 +556,40 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+                for n in range(SET_SIZE):
+                    # Note: This only means SET_SIZE is lower than MAXIMUM_STRATEGIES
+                    assert n < key
+                    assert set[n] != queue[i]
+                    if set[n] == ZERO_ADDRESS:
+                        set[n] = queue[i]
+                        break
+                break
```
fubuloubu, line 585: Doesn't this mean it exhausted the set available space without finding a place for the entry?

fubuloubu (commented):
Overall, I still think there is a potential vulnerability if you attempt to add 2 or more strategies that compute the same key.

Could you please add a test case showing this is not the case? You might have to farm the strategy addresses by nonce farming (keep re-deploying the strategies until you get an address that satisfies this condition), or you can try adding a fuzzer case that tests this logic (e.g. `addStrategy`, `migrateStrategy` and `removeStrategy` as well as all the queue manipulation functions as rules)

contracts/Vault.vy
Niraj-Kamdar: Nope. it will only get triggered when it founds an empty space within the set. other cases will raise an error due to assertions.

Niraj-Kamdar:
SET is designed to handle collisions efficiently as long as `SET_SIZE >= MAXIMUM_STRATEGIES` meets and `SET_SIZE` is power of 2. 

** Why  `SET_SIZE >= MAXIMUM_STRATEGIES`?**
You will need enough space in SET to incorporate maximum number strategies in the worst case scenario.

** Why power of 2 in particular?**
It will be computationally efficient and more collision resistant. 
Ex: `SET_SIZE = 32` => `mask = 31 = b"11111"`, taking `bitwise_and` with the `int` representation of address will give us key. So SET_SIZE with power of 2 will utilise all the bits efficiently. 

If we take `SET_SIZE = 20` then `mask=b"10011"`  => In this case, if `address1 = b"11100"` and `address2=b"10000"` will produce same key and create collision even with 2 different bits but having SET_SIZE as power of 2 length will solve this issue.

** How are you handling collision within the set?
Whenever collision happens, set will search for the next empty slot from the key position linearly. If it won't find it till the end which is the SET_SIZE then it will loop back to the beginning of set and try to find the empty key slot from the 0th index to key position. It will surely find one if `SET_SIZE >= MAXIMUM_STRATEGIES`.

Niraj-Kamdar:
I can add functional test case showing that collision won't be a problem.
Testing collision: 
create 2 addresses with same 5 least significant bits and random 155 most significant bits and convert it to the address.

Niraj-Kamdar:
I will use https://www.ethereumaddressgenerator.com/ for generating addresses with common suffixes.

contracts/Vault.vy
```diff
@@ -553,36 +556,40 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
         assert self.strategies[queue[i]].activation > 0
 
-        existsInOldQueue: bool = False
-        for j in range(MAXIMUM_STRATEGIES):
-            if queue[j] == ZERO_ADDRESS:
-                existsInOldQueue = True
+        # hash will have hash of the address in uint256.
+        # `mask = SET_SIZE - 1` will be used to get key for the set for indexing
+        hash: uint256 = convert(queue[i], uint256)
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `SET_SIZE` bits of address
```
fubuloubu, line 573:
```suggestion
        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `log_2(SET_SIZE)` bits of address
```

contracts/Vault.vy
fubuloubu:
```suggestion
            # We've exhausted the set but still need to place this item
            if m >= SET_SIZE:
```

contracts/Vault.vy
```diff
@@ -553,36 +556,40 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot.
+        for m in range(key, key + SET_SIZE):
+            if m >= SET_SIZE:
+                for n in range(SET_SIZE):
```
fubuloubu, line 578:
```suggestion
                # So we're starting over from the beginning of the set to find space
                for n in range(SET_SIZE):
```

contracts/Vault.vy
fubuloubu: This is not strictly true actually, the `break` internal to the `if` statement above it should trigger if it finds space in the set, this line needs to change to be a `raise` to account for the scenario where no space is found at all
```suggestion
                # NOTE: should not revert if SET_SIZE is lower than MAXIMUM_STRATEGIES
                raise UNREACHABLE  # dev: no space in set found
```

contracts/Vault.vy
fubuloubu, line 580: Delete this line to save a tiny amount of gas now that we raise if the set is filled up prematurely
```suggestion
```

fubuloubu: Yes, this will waste a little extra gas getting to the final conclusion, but this invariant will still be maintained by proxy, and also it reduces costs in the common case workflow of 1-2 passes before filling in the set

fubuloubu (approved):
Some more improvements and documentation needed, then LGTM

fubuloubu:
> I will use https://www.ethereumaddressgenerator.com/ for generating addresses with common suffixes.

Might be more difficult, the address has to be a valid strategy (e.g. has `.vault()` and `.want()`) to be added to the queue. There might be a way to mock this, but it's probably not trivial. I'd recommend the fuzzing case anyways.

Niraj-Kamdar:
No outer `break` shouldn't be replaced with the raise and `assert n < key` is very necessary.

fubuloubu:
> No outer `break` shouldn't be replaced with the raise and `assert n < key` is very necessary.

Please explain, I don't see it

Niraj-Kamdar:
Ex: `key = 30` and both 30 and 31 is filled. Now we start with 0th slot inside the `if m >= SET_SIZE` part, and start looping from start to find empty space if we find the same element at any place we flag it as duplicate and raise an error. if we find an empty slot we put the strategy address there and break the **inner loop**. control goes to the if block outside the inner loop which is inside another loop which you need to break and register the strategy in withdrawal queue.

If you raise an error there a valid transaction would revert.

Niraj-Kamdar:
> Might be more difficult, the address has to be a valid strategy (e.g. has `.vault()` and `.want()`) to be added to the queue. There might be a way to mock this, but it's probably not trivial. I'd recommend the fuzzing case anyways.

I don't know how to do it.  can you help me?

contracts/Vault.vy
fubuloubu: > Ex: `key = 30` and both 30 and 31 is filled. Now we start with 0th slot inside the `if m >= SET_SIZE` part, and start looping from start to find empty space if we find the same element at any place we flag it as duplicate and raise an error. if we find an empty slot we put the strategy address there and break the **inner loop**. control goes to the `if block` outside the inner loop which is inside another loop which you need to break and register the strategy in withdrawal queue.
> 
> If you raise an error there a valid transaction would revert.

Ah, I see the difference now.

The change would be this:
```suggestion
                    # NOTE: should not revert if SET_SIZE is lower than MAXIMUM_STRATEGIES
                    raise UNREACHABLE  # dev: no space in set found

                # NOTE: break out of top loop
                break
```
With this change, removing `assert n < key` above would be valid as that invariant check happens here.

---

Alternatively, one could just `m %= SET_SIZE` on the first line to re-align the set index to be within the loop's boundaries

fubuloubu (commented):
Responded inside the relevant conversation

fubuloubu:
> > Might be more difficult, the address has to be a valid strategy (e.g. has `.vault()` and `.want()`) to be added to the queue. There might be a way to mock this, but it's probably not trivial. I'd recommend the fuzzing case anyways.
> 
> I don't know how to do it. can you help me?

The fuzzing case would be similar to others in the `tests/integration/` section. This section of Brownie's docs will help: https://eth-brownie.readthedocs.io/en/stable/tests-hypothesis-stateful.html

---

Alternatively, you can just keep deploying until you get two strategies with a key of 31. That would test the boundary condition. Add all the intermediate strategies in there (up to 20) for funsies.

Niraj-Kamdar:
No,
```python
                break               
 # NOTE: should not revert if SET_SIZE is lower than MAXIMUM_STRATEGIES                
raise UNREACHABLE  # dev: no space in set found
```
Above would be wrong as well.

Idea of using modulo seems interesting though.

fubuloubu:
> No, replacing `break` with this...
> 
> ```python
>  # NOTE: should not revert if SET_SIZE is lower than MAXIMUM_STRATEGIES                
> raise UNREACHABLE  # dev: no space in set found
> ```
> 
> would be wrong as well.

It would work in context. It should *always* break out of the inner loop, unless the set is full (which it shouldn't be based on the parameterization, but... the `raise` catches that), and then the next break statement throws you out of the outer loop as well, once you break the inner loop.

fubuloubu:
> Idea of using modulo seems interesting though.

The modulo gets you over having the do the inner and outer loop setup

Niraj-Kamdar:
> The modulo gets you over having the do the inner and outer loop setup
Yes but it seems using modulo is more gas expensive because of memory allocation and computation then just using `if` statement.

Niraj-Kamdar:
> It would work in context. It should _always_ break out of the inner loop, unless the set is full (which it shouldn't be based on the parameterization, but... the `raise` catches that), and then the next break statement throws you out of the outer loop as well, once you break the inner loop.

assert at the beginning of inner loop would catch that and if we keep `raise ` in the end it will stop the loop in one round if it can't find an element. We wouldn't want to stop it until we exhaust space, find duplicate or an empty slot.

fubuloubu:
> > It would work in context. It should _always_ break out of the inner loop, unless the set is full (which it shouldn't be based on the parameterization, but... the `raise` catches that), and then the next break statement throws you out of the outer loop as well, once you break the inner loop.
> 
> assert at the beginning of inner loop would catch that and if we keep `raise ` in the end it will stop the loop in one round if it can't find an element. We wouldn't want to stop it until we exhaust space, find duplicate or an empty slot.

This isn't like Python, it'll stop the whole transaction from progressing. It stopping from the `assert` or from the `raise` doesn't change the outcome, only the amount of gas it uses. Using the `assert` means it will use that much gas on every iteration, changing to the `raise` will only use that much gas in the case where the set is full, which a) should definitely not be the common case, and b) should never happen anyways, due to the parameterization. Plus, it makes the code look a lot cleaner IMO.

fubuloubu:
> > The modulo gets you over having the do the inner and outer loop setup
> 
> Yes but it seems using modulo is more gas expensive because of memory allocation and computation then just using `if` statement.

Do you have numbers that show this? Would think it would be more expensive to do the if statements instead of just doing `m %= SET_SIZE`. Surely the codesize would be smaller from removing the 2nd loop.

Niraj-Kamdar:
> Do you have numbers that show this? Would think it would be more expensive to do the if statements instead of just doing `m %= SET_SIZE`. Surely the codesize would be smaller from removing the 2nd loop.

Well `m` will be immutable so we need to declare new variable and assign it `m % SET_SIZE`. i think that's why it's expensive?

fubuloubu:
> > Do you have numbers that show this? Would think it would be more expensive to do the if statements instead of just doing `m %= SET_SIZE`. Surely the codesize would be smaller from removing the 2nd loop.
> 
> Well `m` will be immutable so we need to declare new variable and assign it `m % SET_SIZE`. i think that's why it's expensive?
> 
> These are the numbers with `if`:
> 
> ```
>    ├─ setWithdrawalQueue              -  avg:   66567  avg (confirmed):   93818  low:   28438  high:  179829
> ```
> 
> and these with `%`:
> 
> ```
>    ├─ setWithdrawalQueue              -  avg:   66597  avg (confirmed):   93855  low:   28438  high:  179921
> ```
> 
> Not that much but still depending on gas price it can become expensive

Yeah, that's basically nothing lol. Can you post a diff with this change in play? I think the increased readability is worth the extra 43+ gas spent.

Niraj-Kamdar:
> Yeah, that's basically nothing lol. Can you post a diff with this change in play? I think the increased readability is worth the extra 43+ gas spent.
Yeah it looks much nicer

contracts/Vault.vy
```diff
@@ -553,36 +556,32 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
-                existsInOldQueue = True
+        # hash will have hash of the address in uint256.
+        # `mask = SET_SIZE - 1` will be used to get key for the set for indexing
+        hash: uint256 = convert(queue[i], uint256)
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `log_2(SET_SIZE)` bits of address
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot.
+        for j in range(key, key + SET_SIZE):
+            idx: uint256 = j % SET_SIZE  # traverse set for empty space in circular manner
+            assert set[idx] != queue[i]  # duplicate check
```
fubuloubu, line 578: This notation is specific to brownie
```suggestion
            assert set[idx] != queue[i]  # dev: duplicate in set
```

fubuloubu (approved):
This looks so much better!

contracts/Vault.vy
fubuloubu: With this notation, you can do `with brownie.reverts("dev: duplicate in set"): vault.addStrategy(...)` to make sure it fails on this specific line when adding a duplicated strategy

contracts/Vault.vy
```diff
@@ -553,36 +556,32 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
-        for j in range(MAXIMUM_STRATEGIES):
-            if queue[j] == ZERO_ADDRESS:
-                existsInOldQueue = True
+        # hash will have hash of the address in uint256.
+        # `mask = SET_SIZE - 1` will be used to get key for the set for indexing
+        hash: uint256 = convert(queue[i], uint256)
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  # Key is first `log_2(SET_SIZE)` bits of address
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot.
+        for j in range(key, key + SET_SIZE):
```
fubuloubu, line 576:
```suggestion
        # NOTE: `key` is first `log_2(SET_SIZE)` bits of address (which is a hash)
        hash: uint256 = convert(queue[i], uint256)
        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  
        # Most of the times following for loop only run once which is making it highly gas efficient 
        # but in the worst case of key collision it will run linearly and find first empty slot in the set.
        for j in range(key, key + SET_SIZE):
```

contracts/Vault.vy
```diff
@@ -553,36 +556,32 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
+            idx: uint256 = j % SET_SIZE  # traverse set for empty space in circular manner
```
fubuloubu, line 577:
```suggestion
            # NOTE: we can always find space by treating set as circular (as long as `SET_SIZE >= MAXIMUM_STRATEGIES`)
            idx: uint256 = j % SET_SIZE
```

fubuloubu (commented):
Clean up docs a bit more

contracts/Vault.vy
```diff
@@ -553,36 +556,32 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
 
         assert self.strategies[queue[i]].activation > 0
 
-        existsInOldQueue: bool = False
-        for j in range(MAXIMUM_STRATEGIES):
-            if queue[j] == ZERO_ADDRESS:
-                existsInOldQueue = True
+        # NOTE: `key` is first `log_2(SET_SIZE)` bits of address (which is a hash)
+        hash: uint256 = convert(queue[i], uint256)
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)
```
fubuloubu, line 572: Don't really need the intermediate step here, save some gas
```suggestion
        key: uint256 = bitwise_and(convert(queue[i], uint256), SET_SIZE - 1)  
```

contracts/Vault.vy
```diff
@@ -553,36 +556,32 @@ def setWithdrawalQueue(queue: address[MAXIMUM_STRATEGIES]):
...
-            if queue[j] == ZERO_ADDRESS:
-                existsInOldQueue = True
+        # NOTE: `key` is first `log_2(SET_SIZE)` bits of address (which is a hash)
+        hash: uint256 = convert(queue[i], uint256)
+        key: uint256 = bitwise_and(hash, SET_SIZE - 1)  
+        # Most of the times following for loop only run once which is making it highly gas efficient 
+        # but in the worst case of key collision it will run linearly and find first empty slot in the set.
+        for j in range(key, key + SET_SIZE):
+            # NOTE: we can always find space by treating set as circular (as long as `SET_SIZE >= MAXIMUM_STRATEGIES`)
+            idx: uint256 = j % SET_SIZE
```
fubuloubu, line 577: Curious if this might be a little nicer
```suggestion
        for j in range(SET_SIZE):
            # NOTE: we can always find space by treating set as circular (as long as `SET_SIZE >= MAXIMUM_STRATEGIES`)
            idx: uint256 = (key + j) % SET_SIZE
```

fubuloubu (commented):
Sorry, keep having optimization ideas here

fubuloubu (approved):
LGTM

fubuloubu:
Note: because of the reduction in complexity, this change seemed like value-add.

Still should add a fuzzing test case for queue management that ensures there's no way to violate the invariants in `setWithdrawalQueue`

PR #408, opened 2021-06-15
feat: add strategy registry

steffenix:


contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,70 @@
...
+
+interface Strategy:
+    def name() -> String[64]: view
+
+struct StrategyInfo:
+    name: String[64]
+    numReleases: uint256 
+    versions: address[20]
+
+strategiesReleases: public(HashMap[bytes32, StrategyInfo])
```
fubuloubu, line 17: `String[64]` can be a key. Saves a hash

fubuloubu: Also, not really sure that using the strategy's name is all that safe, but I guess depends on who can add it and what the outcomes are if it's incorrect.

fubuloubu, line 17:
```suggestion
# Strategy.name => release_id
strategiesNumReleases: public(HashMap[String[64], uint256])
# Strategy.name => release_id => Strategy.address
strategyVersions: public(HashMap[String[64], HashMap[uint256, address]])
```

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,70 @@
...
+    strategyHash: bytes32 = sha256(strategyName)
+    if self.strategiesReleases[strategyHash].numReleases == 0:
+	    self.strategiesReleases[strategyHash] = StrategyInfo({
+		    name: strategyName,
+		    numReleases: 0,
+		    versions: [ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS]
+	    })
+    numReleases: uint256 = self.strategiesReleases[strategyHash].numReleases
+    self.strategiesReleases[strategyHash].numReleases = numReleases + 1
+    self.strategiesReleases[strategyHash].versions[numReleases] = strategy
```
fubuloubu, line 64: what's the usecase for tracking multiple releases of the same strategy name?

contracts/StrategyRegistry.vy
poolpitako: I think it's fine, it can be based on github repo. For example:
`yearnV2-generic-lender-strat` for https://github.com/Grandthrax/yearnV2-generic-lender-strat
`SingleSidedCurve` for https://github.com/Grandthrax/SingleSidedCurve

I guess we can also normalize repo's names

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,70 @@
...
+    assert msg.sender == self.pendingGovernance  # dev: unauthorized
+    self.governance = msg.sender
+    log NewGovernance(msg.sender)
+
+@external
+def addNewRelease(strategy :address, name: String[64] = ""):
+    """
+    @notice add a new version of a strategy
+    """
+    assert msg.sender == self.pendingGovernance  # dev: unauthorized
```
poolpitako, line 51: why `pendingGovernance`?

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,70 @@
...
+    strategyName: String[64] = name 
+    if name == "":
+        strategyName = Strategy(strategy).name()
```
poolpitako, line 54: tbh I would revert.

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,70 @@
...
+    strategyHash: bytes32 = sha256(strategyName)
+    if self.strategiesReleases[strategyHash].numReleases == 0:
+	    self.strategiesReleases[strategyHash] = StrategyInfo({
+		    name: strategyName,
+		    numReleases: 0,
+		    versions: [ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS, ZERO_ADDRESS]
```
poolpitako, line 60: this is how the registry works? can we have a map inside the map with key = `numReleases`?

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,70 @@
...
+	    })
+    numReleases: uint256 = self.strategiesReleases[strategyHash].numReleases
+    self.strategiesReleases[strategyHash].numReleases = numReleases + 1
+    self.strategiesReleases[strategyHash].versions[numReleases] = strategy
+
```
poolpitako, line 65: Not sure how we are going to handle the cloning inside here since the parameters will be different.
Have you thought of that issue?

contracts/StrategyRegistry.vy
steffenix: What about different versions of the vault, should the name includes the vault version?

contracts/StrategyRegistry.vy
steffenix: ok

contracts/StrategyRegistry.vy
steffenix: I don't think that is necessary. Find the strategy latest version and then use the clone function from the strategy

contracts/StrategyRegistry.vy
steffenix: the issue is we can't have a hashmap within a structure but @fubuloubu suggested an other structure.

contracts/StrategyRegistry.vy
steffenix: Copied from Registry, changing governance is 2 phases.

contracts/StrategyRegistry.vy
steffenix: So I tough that should work like the registry and keep multiple versions for the same name.
What we need is to handle multiple vault versions for the same strategy name, this can be handled by adding the vault version to the name of the strategy or we save it as part of the hashmap

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,61 @@
...
+@external
+def setGovernance(governance: address):
+    """
+    @notice Starts the 1st phase of the governance transfer.
+    @dev Throws if the caller is not current governance.
+    @param governance The next governance address
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+    self.pendingGovernance = governance
```
salazarguille, line 31: I would emit an event when the `pendingGovernance` variable is changed, so we can track it offchain with thegraph.

steffenix: So we have `log NewGovernance(msg.sender)`, what we have been doing with vaults is only emit when governance is accepted. Do you think `pendingGovernance` would also be a good thing to have  on the vault?

It seems to me it is.

salazarguille: Yes, I think it would be good to have.

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,61 @@
...
+    """
+    @notice add a new version of a strategy
+    """
+    assert msg.sender == self.pendingGovernance  # dev: unauthorized
+    strategyName: String[64] = name 
+    if name == "":
+        strategyName = Strategy(strategy).name()
+        assert strategyName != ""
+    apiVersion: String[8] = Strategy(strategy).apiVersion()
+    self.strategyVersions[StrategyInfo({name: strategyName, apiVersion: apiVersion})] = strategy
```
salazarguille, line 57: Should we emit an event here with strategy name, address, and apiVersion?

contracts/StrategyRegistry.vy
fubuloubu: If we feel strongly that a string is the best domain separator to use here, then yeah this might make sense to do. The norms around "how" the string is set can be decided off-chain, and potentially evolve over time (with some discipline, of course)

contracts/StrategyRegistry.vy
fubuloubu: Don't forget that some of this information might already exist in the Vault registry

contracts/StrategyRegistry.vy
poolpitako: this is for adding a new version of the strategy should be asserting governance not pending.

contracts/StrategyRegistry.vy
poolpitako: well, if we are doing a registry I would like the registry to have the cloning logic.
Perhaps a factory contract? I am trying to trim as much fat as possible from strategies.

salazarguille: I think the complicated part is that each strategy has its own params for the cloning logic. Isn't it?

contracts/StrategyRegistry.vy
```diff
@@ -0,0 +1,61 @@
...
+
+interface Strategy:
+    def name() -> String[64]: view
+    def apiVersion() -> String[8]: view
+
+struct StrategyInfo:
+  name: String[64]
+  apiVersion: String[8]
+
+strategyVersions: public(HashMap[StrategyInfo, address])
```
salazarguille, line 17: AFAIK, to query this mapping we will always need the name and apiVersion to get the strategy address. Right?

steffenix: Yes

contracts/StrategyRegistry.vy
steffenix: got it!

contracts/StrategyRegistry.vy
poolpitako: yep, that's why I was thinking that perhaps the registry can map to a factory.
Something like: `registry.getFactory(likeStrategy)`

salazarguille: We also might decode the params and use a `bytes data` as an input param in the `cloneStrategy` function. For example, using a struct we might

```MyCustomStruct memory mcs = abi.decode(data, (MyCustomStruct));````

Comment (salazarguille) [reply contracts/StrategyRegistry.vy:58]: Just an idea: 
In case a strategy is clonable:
- We call the factory.cloneStrategy(...)
- Factory call the stategy.clone(data) function.
- The strategy should decode the data bytes.
- Clone the strategy.
- Initialize the cloned strategy with the data.
- Finally, the factory should register the new cloned strategy

Comment (steffenix) [reply contracts/StrategyRegistry.vy:58]: So here I want to have two different registry, one that takes care of what's the most up to date version of a strategy (this contract) and an other one that can take care of keeping a list of all strategies (an other contract).

So the cloneStrategy should only be called by governance and then we can have a cloneStrategyExperiment function that can be done by anyone.

File: contracts/StrategyRegistry.vy (hunk 7 above)
Comment (steffenix) [reply contracts/StrategyRegistry.vy:31]: https://github.com/yearn/yearn-vaults/pull/412

File: contracts/StrategyRegistry.vy (hunk 2 above)
Comment (steffenix) [reply contracts/StrategyRegistry.vy]: Can't do that because Dictionary keys must be a base type.

Comment (fubuloubu) [reply contracts/StrategyRegistry.vy]: ah well, `base32` should work.

perhaps should add the ability for any type that can be hashed to be used as a key, since that's really the requirement here.

**Comment by steffenix** [comment]
Since last reviews:

- Added clone method to the StrategyVersionRegistry that slims down the BaseStrategy.
- Prevent clones of clones
- Removed initialized without params (can't find a way to do `abi.encoded(msg.sender,...)`)

File: contracts/StrategyVersionRegistry.vy [hunk 10]
```diff
@@ -0,0 +1,85 @@
...
+    name: String[64]
+    apiVersion: String[8]
+
+event Cloned:
+    clone: address
+
+interface Strategy:
+    def name() -> String[64]: view
+    def apiVersion() -> String[8]: view
+    def initialize(params: Bytes[128]): nonpayable
```
fubuloubu, line 22: You can make this larger for future use (this is a dynamic variable, so this means "up to 256 bytes")
```suggestion
    def initialize(params: Bytes[256]): nonpayable
```

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,85 @@
...
+    @notice add a new version of a strategy
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+    assert Strategy(strategy).isOriginal() # dev: not original 
+    strategyName: String[64] = name 
+    if name == "":
+        strategyName = Strategy(strategy).name()
+        assert strategyName != ""
+    apiVersion: String[8] = Strategy(strategy).apiVersion()
+    key: String[73] = concat(strategyName, SEPARATOR, apiVersion)
```
fubuloubu, line 68: Might be better to abstract this to it's own internal function
```python
@pure
@internal
def _computeKey(name: String[64], version: String[8]) -> String[73]:
    return concat(name, "@", version)  # NOTE: No need for constant here
```

And then just use it via:
```suggestion
    key: String[73] = self._computeKey(strategyName, apiVersion)
```

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,91 @@
...
+    log StrategyRegistered(strategy, strategyName, apiVersion)
+
+@view 
+@external
+def latestRelease(name: String[64], apiVersion: String[8]) -> address:
+    key: String[73] = self._computeKey(name, apiVersion)
+    return self.strategyVersions[key]
+
+@external
+def clone(strategy: address, params: Bytes[128]):
```
fubuloubu, line 83: To match above change
```suggestion
def clone(strategy: address, params: Bytes[256]):
```

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+    assert Strategy(strategy).isOriginal() # dev: not original 
+    strategyName: String[64] = name 
+    if name == "":
+        strategyName = Strategy(strategy).name()
+        assert strategyName != ""
+    apiVersion: String[8] = Strategy(strategy).apiVersion()
+    key: String[73] = self._computeKey(strategyName, apiVersion)
+    self.strategyVersions[key] = strategy
```
jmonteer, line 73: does it make sense to check if it is already set? 

thinking about issues that may come from overwriting a strategy@version with a different deployment

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+
+@view 
+@external
+def latestRelease(name: String[64], apiVersion: String[8]) -> address:
+    key: String[73] = self._computeKey(name, apiVersion)
+    return self.strategyVersions[key]
+
+@external
+def clone(strategy: address, params: Bytes[256]):
+    assert Strategy(strategy).isOriginal()  # dev: not original
```
jmonteer, line 84: should we add this method in BaseStrategy?

variable isOriginal set at deployment time so strategists don't need to handle this

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+
+    newStrategy: address = create_forwarder_to(strategy)
+ 
+    Strategy(newStrategy).initialize(params)
+
+    log Cloned(newStrategy)
```
jmonteer, line 90: as this is a registry, would it make sense to emit more info about original strat? 

original address
original name ? 
original version ?

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+pendingGovernance: public(address)
+
+event NewGovernance:
+    governance: address
+
+event NewPendingGovernance:
+    governance: address
+
+event StrategyRegistered:
+    strategy: address
```
jmonteer, line 12: would it make sense to add this as indexed arg?

contracts/StrategyVersionRegistry.vy
steffenix: It's expected to be overwritten as new version of a strategy are released.

contracts/StrategyVersionRegistry.vy
steffenix: Instead of BaseStrategyInitializable ?
The strategy needs at least to be `Initializable` so I will keep it within the BaseStrategyInitializable.

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+
+@external
+def addNewRelease(strategy :address, name: String[64] = ""):
+    """
+    @notice add a new version of a strategy
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+    assert Strategy(strategy).isOriginal() # dev: not original 
+    strategyName: String[64] = name 
+    if name == "":
```
jmonteer, line 68: Question: why checking if a newly initialized variable is empty? shouldn't it be in 100% of cases? 

in my ignorance, I would directly do:
```
strategyName = Strategy(strategy).name()
assert strategyName != ""
```

contracts/StrategyVersionRegistry.vy
steffenix: cc @salazarguille

contracts/StrategyVersionRegistry.vy
steffenix: name is a parameter:
You can set it to override the strategy name otherwise it will default to the Strategy.name()

```
addNewRelease(strategy :address, name: String[64] = ""):
```

contracts/StrategyVersionRegistry.vy
jmonteer: oh sure, I misinterpreted apiVersion for strategyVersion when writing this

still, any issues that may come from releasing a new version for the same StrategyName@ApiVersion pair? 

like getting all the existing clones from events and having 2 different deployments for same pair?

contracts/StrategyVersionRegistry.vy
steffenix: cc @salazarguille

contracts/StrategyVersionRegistry.vy
steffenix: Sorry didn't got it

contracts/StrategyVersionRegistry.vy
jmonteer: I missed that it is already in BaseStrategyInitializable 

all good

contracts/StrategyVersionRegistry.vy
jmonteer: trying to think about potential issues that may arise from the following process
I release a strategy for yvDAI 0.4.2 
I add it to the registry StrategyA@0.4.2
Then I find a bug in that strategy and redeploy
I need to add it to the registry, the key will still be StrategyA@0.4.2 and it will override previous strategy

might this cause any issue ? 
cc @salazarguille

steffenix: That's the point is you keep updating the strategy address.

steffenix:
Should BaseStrategy and BaseStrategyInitializable be merged together?

jmonteer:
@poolpitako had an opinion on this.

steffenix:
@poolpitako ^

PR #409, opened 2021-06-15
docs: document maxLoss param

### LINKED ISSUE #216: Clarify `maxLoss` parameter in documentation
From audit:

"We would recommend to document a bit more clearly what “maxLoss” means for the withdraw function. We first thought that the accepted loss percentage is computed based on the returned token amount, but it is computed based on the burnt share amount. That is fine, it could just be a bit more clear, but we guess that the UI will take care of that."

steffenix:
closes #216

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1036,6 +1036,7 @@ def withdraw(
         caller's address.
     @param maxLoss
         The maximum acceptable loss to sustain on withdrawal. Defaults to 0.01%.
+        The acceptable amount of token shares to be brunt to cover losses. 
     @return The quantity of tokens redeemed for `_shares`.
     """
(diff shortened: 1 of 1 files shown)


contracts/Vault.vy
```diff
@@ -1036,6 +1036,7 @@ def withdraw(
         caller's address.
     @param maxLoss
         The maximum acceptable loss to sustain on withdrawal. Defaults to 0.01%.
+        The acceptable amount of token shares to be brunt to cover losses.
```
fubuloubu, line 1039:
```suggestion
        If a loss is specified, up to that amount of shares may be burnt to cover losses on withdrawal.
```

fubuloubu (approved):

PR #415, opened 2021-06-22
fix: issue #414

### LINKED ISSUE #414: Can't call setEmergencyExit after calling revokeStrategy
Use case:

I call `revokeStrategy()` so strategy returns funds and after not getting everything I call `setEmergencyExit()` to get everything out.

poolpitako:
Fixes #414

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -1421,8 +1421,10 @@ def revokeStrategy(strategy: address = msg.sender):
     @param strategy The Strategy to revoke.
     """
     assert msg.sender in [strategy, self.governance, self.guardian]
+    # NOTE: This function may be called via `BaseStrategy.setEmergencyExit` while the
+    #       strategy might have already been revoked or had the debt limit set to zero
     if self.strategies[strategy].debtRatio == 0:
-        return # no-op
+        return # already set to zero, nothing to do
 
     self._revokeStrategy(strategy)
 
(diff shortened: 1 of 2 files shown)


fubuloubu:
It looks like you did some swaps of tabs and spaces

poolpitako:
> It looks like you did some swaps of tabs and spaces

you can add `w=1` to the diff and it will remove whitespace diff.

steffenix:
I think we shouldn't commit space changes.

steffenix (approved):

contracts/Vault.vy
```diff
@@ -1421,7 +1421,8 @@ def revokeStrategy(strategy: address = msg.sender):
     @param strategy The Strategy to revoke.
     """
     assert msg.sender in [strategy, self.governance, self.guardian]
-    assert self.strategies[strategy].debtRatio != 0 # dev: already zero
+    if self.strategies[strategy].debtRatio == 0:
+        return # no-op
```
fubuloubu, line 1425:
```suggestion
    # NOTE: This function may be called via `BaseStrategy.setEmergencyExit` while the
    #       strategy might have already been revoked or had the debt limit set to zero
    if self.strategies[strategy].debtRatio == 0:
        return # already set to zero, nothing to do
```

fubuloubu (approved):
All the whitespace changes made this a little harder to review. Next time please put formatting updates in another PR

fubuloubu (approved):

PR #419, opened 2021-06-24
Removed excess strings from require functions to reduce contract size

Tylryan:
Hope this helps!


steffenix:
Only remove the message from the onlyX requires we should keep the others

PR #422, opened 2021-06-25
refactor: reduce base strategy size

### LINKED ISSUE #413: add missing useful modifier to base Strategy onlyManagement
we are seeing more and more uses for a special modifier where only managament + governance can adjust settings,
we should add it to BaseStrategy.sol

```
   modifier onlyManagement() {
        require(
            msg.sender == governance() || msg.sender == vault.management(),
            "!authorized"
        );
        _;
    }
    ```

### PR DESCRIPTION
**Pull request by steffenix**
closes #413 #417 #418

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -272,39 +272,34 @@ abstract contract BaseStrategy {
     bool public emergencyExit;
 
     // modifiers
-    modifier onlyAuthorized() {
-        require(msg.sender == strategist || msg.sender == governance(), "!authorized");
-        _;
+    function _onlyAuthorized() internal {
+        require(msg.sender == strategist || msg.sender == governance());
     }
 
-    modifier onlyEmergencyAuthorized() {
-        require(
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: contracts/BaseStrategy.sol [hunk 1]
```diff
@@ -272,39 +272,34 @@ abstract contract BaseStrategy {
...
                 msg.sender == governance() ||
                 msg.sender == vault.guardian() ||
-                msg.sender == vault.management(),
-            "!authorized"
+                msg.sender == vault.management()
         );
-        _;
+    }
+
+    function onlyManagement() internal {
```
poolpitako, line 301: if you are getting rid of all modifiers, kill this one as well.

poolpitako (approved):
Did you test the final reduction?

contracts/BaseStrategy.sol
steffenix: That's a request from https://github.com/yearn/yearn-vaults/issues/413 but can only be added by strategist.

PR #425, opened 2021-06-28
fix: shares for amount calculation should use freeFunds

### LINKED ISSUE #390: Question about recalculation of shares in withdraw
Should the recalculation of shares to burn on withdraw use `freeFunds` instead of `totalAssets`?

First  calculation of `shares` to `value` uses `freeFunds`
https://github.com/yearn/yearn-vaults/blob/master/contracts/Vault.vy#L1048

Second calculation of `value + totalLoss` to `shares` uses `totalAssets`

https://github.com/yearn/yearn-vaults/blob/master/contracts/Vault.vy#L1101

Share holder loses less share when second calculation is performed with `totalAssets`

steffenix:
fixes #390

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -813,6 +813,11 @@ def _calculateLockedProfit() -> uint256:
     else:
         return 0
 
+@view
+@internal
+def _freeFunds() -> uint256:
+    return self._totalAssets() - self._calculateLockedProfit()
+
 @internal
 def _issueSharesForAmount(to: address, amount: uint256) -> uint256:
     # Issues `amount` Vault shares to `to`.
@@ -825,8 +830,7 @@ def _issueSharesForAmount(to: address, amount: uint256) -> uint256:
     if totalSupply > 0:
         # Mint amount of shares based on what the Vault is managing overall
         # NOTE: if sqrt(token.totalSupply()) > 1e39, this could potentially revert
-        freeFunds: uint256 = self._totalAssets() - self._calculateLockedProfit()
-        shares =  amount * totalSupply / freeFunds  # dev: no free funds
+        shares =  amount * totalSupply / self._freeFunds()  # dev: no free funds
     else:
         # No existing shares, so mint 1:1
         shares = amount
@@ -913,11 +917,10 @@ def _shareValue(shares: uint256) -> uint256:
 
     # Determines the current value of `shares`.
     # NOTE: if sqrt(Vault.totalAssets()) >>> 1e39, this could potentially revert
-    freeFunds: uint256 = self._totalAssets() - self._calculateLockedProfit()
 
     return (
         shares
-        * freeFunds
+        * self._freeFunds()
         / self.totalSupply
     )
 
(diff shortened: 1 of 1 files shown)


contracts/Vault.vy
```diff
@@ -813,6 +813,11 @@ def _calculateLockedProfit() -> uint256:
     else:
         return 0
 
+@view
+@internal
+def freeFunds() -> uint256:
```
jmonteer, line 818: minor: I would use `_freeFunds` to be consistent with how the rest of internal view functions have been named

jmonteer (commented):
added a comment for naming

otherwise LGTM

contracts/Vault.vy
```diff
@@ -825,8 +830,7 @@ def _issueSharesForAmount(to: address, amount: uint256) -> uint256:
     if totalSupply > 0:
         # Mint amount of shares based on what the Vault is managing overall
         # NOTE: if sqrt(token.totalSupply()) > 1e39, this could potentially revert
-        freeFunds: uint256 = self._totalAssets() - self._calculateLockedProfit()
-        shares =  amount * totalSupply / freeFunds  # dev: no free funds
+        shares =  amount * totalSupply / self.freeFunds()  # dev: no free funds
```
fubuloubu, line 833:
```suggestion
        shares =  amount * totalSupply / self._freeFunds()  # dev: no free funds
```

contracts/Vault.vy
```diff
@@ -913,11 +917,10 @@ def _shareValue(shares: uint256) -> uint256:
 
     # Determines the current value of `shares`.
     # NOTE: if sqrt(Vault.totalAssets()) >>> 1e39, this could potentially revert
-    freeFunds: uint256 = self._totalAssets() - self._calculateLockedProfit()
 
     return (
         shares
-        * freeFunds
+        * self.freeFunds()
```
fubuloubu, line 923:
```suggestion
        * self._freeFunds()
```

contracts/Vault.vy
```diff
@@ -932,7 +935,7 @@ def _sharesForAmount(amount: uint256) -> uint256:
         return  (
             amount
             * self.totalSupply
-            / self._totalAssets()
+            / self.freeFunds()
```
fubuloubu, line 938:
```suggestion
            / self._freeFunds()
```

fubuloubu, line 938: Why this change?

steffenix: Because the shares for the amount in the event of a loss should be calculated based on the free funds and not the totalAssets https://github.com/yearn/yearn-vaults/issues/390

contracts/Vault.vy
steffenix: Thanks for pointing this out made a change to the function name but not the call 🤦‍♂️

contracts/Vault.vy
fubuloubu: Okay. Should probably also adjust the `if` statement above (and cache that value too)

steffenix: 👌

fubuloubu (approved):

fubuloubu:
Should a test be added to cacth this behavior?

steffenix:
> Should a test be added to cacth this behavior?

Should we add a function that exposes `_sharesForAmount` similar to `_shareValue` that's exposed by `pricePerShare`

fubuloubu:
> > Should a test be added to cacth this behavior?
> 
> Should we add a function that exposes `_sharesForAmount` similar to `_shareValue` that's exposed by `pricePerShare`
> 
> Otherwise, the test will need to be performed on a withdraw
> 

On withdrawal sounds appropriate

jmonteer (approved):

PR #427, opened 2021-07-01
chore: update js deps

steffenix:
Not much here.


fubuloubu (approved):
Would prefer to remove JS completely lol

PR #428, opened 2021-07-01
fix: allow harvest on add strategy block

steffenix:


contracts/Vault.vy
```diff
@@ -1634,6 +1634,9 @@ def _assessFees(strategy: address, gain: uint256) -> uint256:
     # Issue new shares to cover fees
     # NOTE: In effect, this reduces overall share price by the combined fee
     # NOTE: may throw if Vault.totalAssets() > 1e64, or not called for more than a year
+    if self.strategies[strategy].activation == block.timestamp:
+        return 0
```
fubuloubu, line 1638:
```suggestion
        return 0  # NOTE: Just added, no fees to assess
```

fubuloubu (approved):
Nice change!

PR #429, opened 2021-07-01
fix: save gas on DEPOSIT_EVERYTHING not re approving

steffenix:
When depositing everything, the safeApprove is triggered every time, this will reduce the gas usage.

### DIFF
--- a/contracts/BaseWrapper.sol
+++ b/contracts/BaseWrapper.sol
@@ -153,11 +153,10 @@ abstract contract BaseWrapper {
         VaultAPI _bestVault = bestVault();
 
         if (pullFunds) {
-            if (amount != DEPOSIT_EVERYTHING) {
-                SafeERC20.safeTransferFrom(token, depositor, address(this), amount);
-            } else {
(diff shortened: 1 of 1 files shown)


contracts/BaseWrapper.sol
```diff
@@ -153,11 +153,10 @@ abstract contract BaseWrapper {
         VaultAPI _bestVault = bestVault();
 
         if (pullFunds) {
-            if (amount != DEPOSIT_EVERYTHING) {
-                SafeERC20.safeTransferFrom(token, depositor, address(this), amount);
-            } else {
-                SafeERC20.safeTransferFrom(token, depositor, address(this), token.balanceOf(depositor));
+            if (amount == DEPOSIT_EVERYTHING) {
+                amount = token.balanceOf(depositor);
```
fubuloubu, line 157: If I understand this change correctly, the update of `amount` here will ensure the `safeApprove` call is skipped on line 163 based on the current allowance

fubuloubu (approved):

contracts/BaseWrapper.sol
steffenix: Yes that's the goal

PR #431, opened 2021-07-02
feat: base router

steffenix:


steffenix:
@fubuloubu I think `migrate` shouldn't be on the BaseRouter and should be kept on BaseWrapper.

fubuloubu:
> @fubuloubu I think `migrate` shouldn't be on the BaseRouter and should be kept on BaseWrapper.
> 

Agree! I also have some other ideas

pandadefi:
https://github.com/yearn/yearn-vaults/pull/443

PR #432, opened 2021-07-03
remove bi-located AffiliateToken

steffenix:


fubuloubu (commented):
You can definitely set this up as a session fixture instead of downloading it like this. Would also be nice to execute the test suite from that repo as well.

stcrab:
> You can definitely set this up as a session fixture instead of downloading it like this. Would also be nice to execute the test suite from that repo as well.



> You can definitely set this up as a session fixture instead of downloading it like this. Would also be nice to execute the test suite from that repo as well.

I have tried doing it before making a script but didn't managed to have it work.

```
@pytest.fixture(scope="session", autouse=True)
def get_affiliate_token():
    url = "https://raw.githubusercontent.com/yearn/brownie-wrapper-mix/master/contracts/AffiliateToken.sol"
    r = requests.get(url, allow_redirects=True)

    open("contracts/test/AffiliateToken.sol", "w").write(
        r.text.replace("@yearnvaults/contracts", "..")
    )
```

fubuloubu:
> > You can definitely set this up as a session fixture instead of downloading it like this. Would also be nice to execute the test suite from that repo as well.
> 
> I have tried doing it before making a script but didn't managed to have it work.

Yeah, it's not going to work because the contract file isn't compiling. You should work around it similar to how we patch the Vault code for working with the API version: https://github.com/yearn/yearn-vaults/blob/2f8db174d061e305538a8f34061c9ab1590d2142/tests/conftest.py#L80

stcrab:
Interesting, seems I still have some issues with some libs.

```
E       brownie.exceptions.CompilerError: solc returned the following errors:
E
E       <stdin>.sol:4:1: ParserError: Source "@openzeppelin/contracts/token/ERC20/ERC20.sol" not found: File outside of allowed directories.
E       import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
```

fubuloubu:
> Interesting, seems I still have some issues with some libs.
> 
> ```
> E       brownie.exceptions.CompilerError: solc returned the following errors:
> E
> E       <stdin>.sol:4:1: ParserError: Source "@openzeppelin/contracts/token/ERC20/ERC20.sol" not found: File outside of allowed directories.
> E       import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
> ```

That's weird, are we using a different alias for OZ in this repo versus that one?

pandadefi:
> That's weird, are we using a different alias for OZ in this repo versus that one?

Alias is the same.

pandadefi:
git push --set-upstream origin remove_file_being_in_both_repo

pandadefi:
https://github.com/yearn/yearn-vaults/pull/442

PR #435, opened 2021-07-06
fix: set default profit and loss ratio

stcrab:


storming0x (approved):
changes default values LGTM! if CI looks green we can merge

PR #439, opened 2021-07-09
chore: update yaml

fubuloubu:
we are updating the `master` branch to `main`


contracts/BaseStrategy.sol
```diff
@@ -922,7 +922,7 @@ abstract contract BaseStrategyInitializable is BaseStrategy {
         address _rewards,
         address _keeper
     ) external returns (address newStrategy) {
-        // Copied from https://github.com/optionality/clone-factory/blob/master/contracts/CloneFactory.sol
+        // Copied from https://github.com/optionality/clone-factory/blob/main/contracts/CloneFactory.sol
```
poolpitako, line 925: this should be reverted

PR #440, opened 2021-07-13
feat: test small losses

poolpitako:
Here's an edge case of small loss not dropping the debt ratio.
The "root issue" is:
 
```
ratio_change: uint256 = min(
            loss * self.debtRatio / self.totalDebt,
            self.strategies[strategy].debtRatio,
        )
```

min there works like `min(int(a), int(b))`. if loss is small enough, `loss * self.debtRatio / self.totalDebt` will be greater than 0 but smaller than 1.


fubuloubu (approved):

pandadefi:
Can you rebase @poolpitako ?

pandadefi:
I have fucked up the rebase so change were lost and the branch got closed.

PR #443, opened 2021-07-14
feat: base router

pandadefi:
Transform the BaseWrapper so that it can handle multiple tokens.
Funds are not held by the router, it acts as a router.
BaseWrapper is rebuilt on top of BaseRouter.


fubuloubu (commented):
Might want to update `_withdraw` and `_deposit` in `BaseWrapper._migrate`

pandadefi:
Can we push this @fubuloubu

fubuloubu (approved):

PR #448, opened 2021-07-15
feat: call health check contract from vault

pandadefi:


contracts/Vault.vy
```diff
@@ -88,6 +88,8 @@ management: public(address)
 guardian: public(address)
 pendingGovernance: address
 
+commonHealthCheck: public(address)
```
storming0x, line 91: healthCheck as name works fine i think

contracts/CommonHealthCheck.sol
```diff
@@ -94,37 +73,42 @@ contract CommonHealthCheck {
     }
 
     function check(
+        address strategy,
```
storming0x, line 76: we should overload this to have two methods called `check` one thats used by vault now, and the old one used in existing strategies, we want to make this the global healthcheck for both vaults and strategies and this new version will become `health.ychad.eth` so we have a single contract to manage this. Both the existing healthchecks and this new vault checks should be supported.

Something to test in the test suite that it can support strategies healthchecks for old versions.

contracts/Vault.vy
```diff
@@ -58,8 +58,8 @@ interface Strategy:
     def migrate(_newStrategy: address): nonpayable
 
 
-interface CustomHealthCheck:
-    def check(profit: uint256, loss: uint256, callerStrategy: address) -> bool: view
+interface CommonHealthCheck:
```
storming0x, line 61: HealtCheck Interface sounds good enough

contracts/Vault.vy
```diff
@@ -1324,26 +1322,9 @@ def setStrategyEnforceChangeLimit(strategy: address, enabled: bool):
...
-def setStrategyCustomCheck(strategy: address, _customCheck: address):
-    """
-    @notice
-        Change the custom strategy, default value is 0x0, when set to a non
-        null address, the vault will check the strategy health using this address.
-        If set to 0x0 it will use default checks.
-    @param strategy The Strategy to update.
-    @param _customCheck The contract that should perform the check, can be set to 0x0.
-    """
+def setCommonHealthCheck(_commonHealthCheck: address):
```
storming0x, line 1325: setHealthCheck, same for arg

contracts/Vault.vy
```diff
@@ -1735,21 +1713,18 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     # Check report is within healty ranges
 
     if self.strategies[msg.sender].enforceChangeLimit:
```
storming0x, line 1715: `if self.healthCheck != ZERO_ADDRESS && HealthCheck(self.healthCheck).doHealthCheck(msg.sender)`

removes the struct modification and delegates the disabling of one time healthcheck to the contract.

Also not ideal but we can entirely disable healthchecks for a vault if there is a problem as a nuclear option by setting to 0x0.

contracts/CommonHealthCheck.sol
```diff
@@ -94,37 +73,42 @@ contract CommonHealthCheck {
...
         uint256 _profit,
         uint256 _loss,
         uint256 _totalDebt
     ) internal view returns (bool) {
-        Limits memory limits = strategiesLimits[msg.sender];
+        Limits memory limits = strategiesLimits[strategy];
         uint256 _profitLimitRatio;
         uint256 _lossLimitRatio;
         if (limits.exists) {
```
storming0x, line 114: we can add a check here based on  a mapping to skip the Next Health check one time, and is setup per strategy address and can be set by authorized. this will be reset back to false after this one time.

That way the disabling of a one time healthcheck logic lies in this contract and saves some space on vault.

storming0x (changes requested):
Good start, Left a few comments and suggestions

contracts/CommonHealthCheck.sol
```diff
@@ -94,37 +73,42 @@ contract CommonHealthCheck {
...
         uint256 loss,
         uint256 debtPayment,
         uint256 debtOutstanding,
         uint256 totalDebt
     ) external view returns (bool) {
-        return _runChecks(profit, loss, debtPayment, debtOutstanding, totalDebt);
+        require(strategy != address(0));
+
+        return _runChecks(strategy, profit, loss, debtPayment, debtOutstanding, totalDebt);
     }
```
storming0x, line 87: we can add a new view method called doHealthCheck(address arg)

that returns true by default unless we have a mapping setter for that address turned to true, in the next harvest for that address it can be disabled as a one time harvest free of health checks.

does it make sense?

pandadefi: I took a similar approach but used a `mapping(address => bool) public disabledCheck;` so that we don't have to have a structure with a `bool exists;` in order to have it enabled by default.

With that said doHealthCheck looks like this.

```
    function doHealthCheck(address _strategy) {
        return !disabledCheck[_strategy];
    }
```

And I had to add a enableCheck that can only be used by the vault that is attached to the strategy

```
    function enableCheck(address _strategy) external onlyVault(_strategy) {
        disabledCheck[_strategy] = false;
    }

```

contracts/CommonHealthCheck.sol
pandadefi: Tested and tagged for reference here:
https://github.com/yearn/yearn-vaults/compare/v0.4.3...v0.4.3-1

contracts/Vault.vy
```diff
@@ -1733,23 +1708,21 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     assert self.token.balanceOf(msg.sender) >= gain + _debtPayment
 
     # Check report is within healty ranges
```
storming0x, line 1710: small typo _healthy_

contracts/Vault.vy
```diff
@@ -1733,23 +1708,21 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     assert self.token.balanceOf(msg.sender) >= gain + _debtPayment
```
storming0x, line 1708: Im thinking of a very deep edge case thats unlikely but nevertheless, it may be an attack.

MAlicious CustomCheck is activated by management that has a deposit balance in vault, 
when we check the numbers here its all good, but then malicious custom check when it gets called in health check proceeds to withdraw balance as reentrancy and strategy is in queue 0 index so it removes the balance reported before, stealing from vault profits or affecting vault accounting.

In that case the the assert invariant in 1708 is no longer valid after healthcheck since balanceOf changed. This is a highly unlikely and difficult attack to pull, but thinking if this assert should be either moved after healthcheck to enforce the balance.

storming0x: Can we add a small test to check if thats possible? Doggie mentioned that if the healthcheck call is VIEW it cannot make state changing calls after in calling stack. @pandadefi can we test with a mock custom check or malicious check that holds balance in the vault?

pandadefi: I have added a check to make sure the call is a STATICCALL

https://github.com/yearn/yearn-vaults/pull/448/commits/7438b23cb6c065c8f0b28b4ef3008c78459a64b5

pandadefi: I have also moved the balance check after the health check

https://github.com/yearn/yearn-vaults/pull/448/commits/5dabf5410730b23e29f4c7dc7cd2a49f514b4841

storming0x (approved):
LGTM!

PR #452, opened 2021-07-20
feat: add strategy registry

pandadefi:


contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+
+@external
+def setGovernance(governance: address):
+    """
+    @notice Starts the 1st phase of the governance transfer.
+    @dev Throws if the caller is not current governance.
+    @param governance The next governance address
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+    log NewPendingGovernance(msg.sender)
```
poolpitako, line 44: `governance` instead of `msg.sender`?

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+def addNewRelease(strategy :address, name: String[64] = ""):
+    """
+    @notice add a new version of a strategy
+    """
+    assert msg.sender == self.governance  # dev: unauthorized
+    assert Strategy(strategy).isOriginal() # dev: not original 
+    strategyName: String[64] = name 
+    if name == "":
+        strategyName = Strategy(strategy).name()
+        assert strategyName != ""
```
poolpitako, line 70: can't you reuse `name`? why do you need `strategyName`?

contracts/StrategyVersionRegistry.vy
```diff
@@ -0,0 +1,90 @@
...
+
+@view 
+@external
+def latestRelease(name: String[64], apiVersion: String[8]) -> address:
+    key: String[73] = self._computeKey(name, apiVersion)
+    return self.strategyVersions[key]
+
+@external
+def clone(strategy: address, params: Bytes[256]):
+    assert Strategy(strategy).isOriginal()  # dev: not original
```
poolpitako, line 84: you are checking that the strategy is original but you are not checking if the strategy belongs to the registry. why?

contracts/StrategyVersionRegistry.vy
pandadefi: That's not possible, I get a `ImmutableViolation: Cannot write to calldata` if I try to reuse name

contracts/StrategyVersionRegistry.vy
pandadefi: The strategy doesn't have to be in the registry to be cloned. Registry helps out finding which strategy is the latest and then you can call the clone function with the address, but you can also clone a strategy that's not part of the registry..

WDYT?

contracts/StrategyVersionRegistry.vy
pandadefi: yes

contracts/StrategyVersionRegistry.vy
poolpitako: interesting!

contracts/StrategyVersionRegistry.vy
poolpitako: I believe only strategies in the registry should be clonable. The same way the vault registry works.

pandadefi: ok, I have made some checks but since you can put a custom name the check can be bypassed.

pandadefi:
closing this for now

PR #460, opened 2021-08-11
feat: tax strategy harvests + add IRS management

0xUncleSam:
Add required IRS integration for yearn strategies. Yearn Finance falls under the US legal definition of a US based corporation and must be taxed on its proceeds. Failure to comply may result in severe financial penalties, repossession of Yearn private key management, and the deanonymization of key Yearn contributors.

This includes a few updates:
- Adding US executive branch as governance and removing the illegal DAO reference
- Allow US executive branch to set IRS address
- Expand IRS role to allow setting tax rate and allow IRS to be a keeper
- Tax each harvest on a configurable tax rate set by the IRS

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -265,6 +265,9 @@ abstract contract BaseStrategy {
     // harvest trigger. See `setDebtThreshold()` for more details.
     uint256 public debtThreshold;
 
+    // Adjustable by the IRS at the whims of our geriatric legislators 
+    uint256 public taxRate;
+
     // See note on `setEmergencyExit()`.
     bool public emergencyExit;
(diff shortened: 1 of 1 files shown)


sdan:
spotted the fed

contracts/BaseStrategy.sol
```diff
@@ -348,6 +357,13 @@ abstract contract BaseStrategy {
         emit UpdatedStrategist(_strategist);
     }
 
+    function setTaxRate(uint256 _taxRate) external {
+        _onlyIrs();
+        require(_taxRate != 0);
+        require(_taxRate <= 100);
```
x48-crypto, line 363: Should this be in bips?

PR #462, opened 2021-08-13
fix: amount needed sanity check

pandadefi:


pandadefi:
@storming0x 
I have got one hedge case that gets triggered here.
When a strategy get funds stolen, 
In the test strategy with a total debt of 200 but only 100 left on the contract.
The user try to withdraw what's left 100.
The TestStrategy reports both 100 of loss and 100 of amountFreed so that's 200 compared to the 100 of _amountNeeded and the check fails.

storming0x:
> @storming0x
> I have got one hedge case that gets triggered here.
> When a strategy get funds stolen,
> In the test strategy with a total debt of 200 but only 100 left on the contract.
> The user try to withdraw what's left 100.
> The TestStrategy reports both 100 of loss and 100 of amountFreed so that's 200 compared to the 100 of _amountNeeded and the check fails.

Yeah i think this is for an inviariant under normal operation, i see your point, but i guess for that edge case we would need to harvest strategy and realized and socialized the loss, that strategy would probably be removed from queue for that exceptional scenario.

But good to think about that.

pandadefi:
An other option is to use a >= instead of == if we care only about the amount + loss being not bellow the amount to withdraw

storming0x:
Also this is not a change needed in main i believe, since vault 0.4.3 handles this already, should this pr go into a branch for tag?

@pandadefi

storming0x:
> An other option is to use a >= instead of == if we care only about the amount + loss being not bellow the amount to withdraw

@Grandthrax @poolpitako @fp-crypto

pandadefi:
The -2 have been tagged already

PR #465, opened 2021-08-26
perf: cache vault balance in withdraw func

t4sk:
replace query of token balance `self.token.balanceOf(self)` with local variable


pandadefi:
Closing this, it will be replaced by https://github.com/yearn/yearn-vaults/pull/482/files

PR #468, opened 2021-09-06
chore: bump vault vyper version

pandadefi:


tests/functional/conftest.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 from brownie import Token, TokenNoReturn
+from eth_account import Account
```
fubuloubu, line 4: What's this import for?

fubuloubu (approved):

tests/functional/conftest.py
pandadefi: removed

fubuloubu:
Is this Vault upgrade going to be released soon?

pandadefi:
We should bring the discussion to the team, since there is some gas saving we can push a new version for future vaults

PR #471, opened 2021-09-29
feat: deprecate lockedProfitDegradation

lbertenasco:
deprecating `lockedProfitDegradation` in favor of a more dynamic approach.

added:
`previousHarvestTimeDelta: public(uint256) # how much time elapsed between last and previous report`

how it works:
main idea was to have a way to evenly distribute pps increase over the same period of time on which those funds were staked, this is to mitigate deposit/harvest/wait-6hs/withdraw attacks.
`previousHarvestTimeDelta` will always carry the biggest delta time. i.e. if a harvest(h2) happens after 10 hours of the previous harvest(h1), then we set `previousHarvestTimeDelta = 10 hours`, but if a new harvests(h3) happens after just hr of h2 `previousHarvestTimeDelta = 8`, and `lockedProfit` is adjusted for the 2 hours that went by.

on the other hand when h1 = t0, h2 = t10 and h3 = t20, `previousHarvestTimeDelta = 10`
same as when h1 = t0, h2 = t10 and h3 = t18, `previousHarvestTimeDelta = 8`

this new approach plays well with ySwaps' async airdrops on between harvests


contracts/Vault.vy
```diff
@@ -1774,6 +1761,13 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
 
     # Update reporting time
     self.strategies[msg.sender].lastReport = block.timestamp
+    currentTimeDelta: uint256 = block.timestamp - self.lastReport
+    # maintains longer (fairer) harvest periods on close timed harvests
+    if self.previousHarvestTimeDelta > currentTimeDelta * 2:
```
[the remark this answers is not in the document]
lbertenasco: `* 2` is used to only "carry" the `previousHarvestTimeDelta` difference only if it'd exceed the newer one.
some examples:
`h1 = t0, h2 = t10 and h3 = t14, previousHarvestTimeDelta = 6 , locked until t20`
^ h2 is getting carried into h3 (minus time delta 4) since it was previously trying to reach t20. so it continues to spread the lock up to that point, and thus avoids reducing the previous distribution time.

`h1 = t0, h2 = t10 and h3 = t15, previousHarvestTimeDelta = 5, locked until t20`
^ here it behaves equal in both scenarios, so we actually skip the if to avoid extra storage read and subtraction cost

`h1 = t0, h2 = t10 and h3 = t16, previousHarvestTimeDelta = 6, locked until t22`
`h1 = t0, h2 = t10 and h3 = t17, previousHarvestTimeDelta = 7, locked until t24`
^ here we see what happens if next total time delta is higher than previous period remaining.

pandadefi (approved):
Interesting approach, still some tests to fix

contracts/Vault.vy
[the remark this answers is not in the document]
fubuloubu: This algorithm needs a much longer comment explaining how it works. Is there a way to re-organize the math to make it clearer what is going on? Perhaps:
```python
if self.previousHarvestTimeDelta - currentTimeDelta > currentTimeDelta:
    ...
```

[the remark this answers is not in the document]
lbertenasco: > ```python
> if self.previousHarvestTimeDelta - currentTimeDelta > currentTimeDelta:
>     ...
> ```
I think this will revert if `currentTimeDelta > self.previousHarvestTimeDelta` and brick the vault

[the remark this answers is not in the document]
fubuloubu: fair

contracts/Vault.vy
```diff
@@ -1762,7 +1768,21 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
...
+    #       h1 = t0, h2 = t10 and h3 = t14, previousHarvestTimeDelta = 6 , locked until t20
+    #       on 2nd example: h2 is getting carried into h3 (minus time delta 4) since it was previously trying to reach t20.
+    #       so it continues to spread the lock up to that point, and thus avoids reducing the previous distribution time.
+    #
+    #       if locked-until is unchanged, to avoid extra storage read and subtraction cost
+    #       h1 = t0, h2 = t10 and h3 = t15, previousHarvestTimeDelta = 5, locked until t20
+    #
+    #       if next total time delta is higher than previous period remaining, locked-until will increase
+    #       h1 = t0, h2 = t10 and h3 = t16, previousHarvestTimeDelta = 6, locked until t22
+    #       h1 = t0, h2 = t10 and h3 = t17, previousHarvestTimeDelta = 7, locked until t24
```
fubuloubu, line 1785: Are these numbers right?

fubuloubu, line 1785: In all these examples, talk about the tradeoff of `previousHarvestTimeDelta` (as computed from previous report) and `currentTimeDelta`

lbertenasco: i think so, yeah
`16+6 = 22`
`17+7 = 24`

fubuloubu: I meant `previousHarvestTimeDelta`

contracts/Vault.vy
```diff
@@ -1774,6 +1767,37 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
...
+    #           currentTimeDelta = 7, (new)previousHarvestTimeDelta = 7 locked until t24
+    #       
+    #       currentTimeDelta is the time delta between now and lastReport.
+    #       previousHarvestTimeDelta is the time delta between lastReport and the previous lastReport
+    #       previousHarvestTimeDelta is assigned the higher value between currentTimeDelta and (previousHarvestTimeDelta - currentTimeDelta)
+    currentTimeDelta: uint256 = block.timestamp - self.lastReport
+    if self.previousHarvestTimeDelta > currentTimeDelta * 2:
+        self.previousHarvestTimeDelta = self.previousHarvestTimeDelta - currentTimeDelta
+    else:
+        self.previousHarvestTimeDelta = currentTimeDelta
```
fubuloubu, line 1799: One thing to take into consideration is that if a harvest period is "late" (aka Keep3r issues or something causing a prolonged period without reports), then when it becomes active again it will delay profit much longer than expected.

Not sure if this is a desireable quality, but if not we could "saturate" the timing here by placing a governance-configurable "maximum" on the delay (e.g. 24 hours)

lbertenasco: that should be the desired behavior, since if harvest get delayed, it potentially had a higher impact in PPS, and we'll be, by setting a max delay, creating an incentive for depositors to snipe those delayed-harvests to get unfair increase on their deposits for a lower period.

fubuloubu: Let's say the max delay is 24 hrs, and the profit is $50k because we haven't harvested in a week. That means the fade-in profit would be something like $50k/24hr = $2k/hr, total accross all depositors. I would argue that it is still doing it's job of discouraging sandwiches, but with improved UX of not having to wait another week for full profit to be dispersed (actually 2 whole weeks, wouldn't it?)

At the very least, we can set it to something large enough to be effectively "disabled" but use it as an extra switch in times of need. Just my 2 cents, we are losing a lever of control here (which is arguably a good thing at a high level)

lbertenasco: I'd like to go without the max delay.
in those rare cases where we do have really big delays, subsequent harvests will not push the rewards further back, but they'll increase the total profit rate. (if they were profitable harvests)
if this happens a user can jump in right before the harvest and get 7x rewards on their deposit over 24hs.
We want an incentive for users to stay for as long as possible. delaying profits does not affect APY, it makes it more stable and/or predictable.
It's also one less thing to worry about setting up correctly.
I think tho, that this needs further testing and experimentation to see what will happen on those edge cases. (also when quickly reporting a loss after a big delayed profitable harvest)

skeletor-spaceman (approved):

skeletor-spaceman:
Deprecated for v2, moving improvements to v3 design.

Issue #467, opened 2021-09-01
setRewards() in BaseStrategy should be settable by gov.


pandadefi:
Request from @poolpitako 
```
setRewards() in BaseStrategy should be settable by gov.

```

Todo:
- add the changes.
- cherry-pick the changes on older version and tag a new version

pandadefi:
That's the idea, you will also have to change the tests so that's tested

PR #474, opened 2021-10-14
fix: updated setRewards

jhb10c:
This is update was requested in [467](https://github.com/yearn/yearn-vaults/issues/467).

The contract BaseStrategy has the onlyRewards updated so that only governance can update the address. The corresponding tests have been updated.


pandadefi:
All checks have passed, one last change regarding the comment and I will merge.

Issue #16, opened 2020-10-14
Add linting to CI for Vyper files


chriswessels:
Currently have linting for Python and Solidity files, but not Vyper.

Add linting support + hook up to Github workflows.

spinoch:
I'm looking into creating such a tool. Is there any chance for compensation in case it resolves this issue? (Not asking for specific numbers, but would be nice to get something)

fubuloubu:
@spinoch yes! I think I can find some individuals that can put together a bounty if you are willing to help take this on!

spinoch:
Cool. I'll see if I can come up with something by next week

fameal:
Hey @spinoch! It seems that you've done a lot of progress in the PR. Are you still working on it?

spinoch:
Hey, @fameal . In short, no. We need to fix a few issues in the blackadder repo and I haven't had time lately. I still would like to get the issues fixed but I can't really give a time frame atm

pandadefi:
We need someone to build and maintain a linter. The linter should be able to work with multiple versions of vyper

fubuloubu:
> We need someone to build and maintain a linter. The linter should be able to work with multiple versions of vyper

note that @spinoch made a first cut of it here: https://github.com/vyperlang/blackadder

AbnerZheng:
@fubuloubu I would like try to finish the remaining work. I have looked through the PR(https://github.com/yearn/yearn-vaults/pull/52), the remaining work is to (correctly) reformat code for mutability of the method, am I right?

fubuloubu:
> @fubuloubu I would like try to finish the remaining work. I have looked through the PR(#52), the remaining work is to (correctly) reformat code for mutability of the method, am I right?

No, there is updates needed to the tool for it to work

AbnerZheng:
@fubuloubu Yeah,  I have checked the tool(https://github.com/vyperlang/blackadder), currently it is simply based on regexp, should I modified it according to the AST to make a normal formatter? Or maybe fork the black(https://github.com/psf/black), and modified the code according to the grammar rule of Vyper?

fubuloubu:
> @fubuloubu Yeah,  I have checked the tool(https://github.com/vyperlang/blackadder), currently it is simply based on regexp, should I modified it according to the AST to make a normal formatter? Or maybe fork the black(https://github.com/psf/black), and modified the code according to the grammar rule of Vyper?

Will accept PRs to that repo, yes

[...]

PR #479, opened 2021-10-24
feat: vyper lint

0xbeedao:
This is a forked version of flake8-vyper, updated to work for version 2.16 of Vyper, and with a Github Action to automate the linting.  It is submitted for the open ticket #16 , adding linting for Vyper.


contracts/Registry.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.2.11
+# @version 0.2.16
```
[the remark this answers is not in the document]
0xbeedao: Rolled this back with the new solution.  No need to change version.

0xbeedao:
I have made the linting now get installed via a simple Pip installation, rather than adding that code to this repo, and also made the Github Action for the linter able to handle the two different Vyper versions in use.

Note that this will never pass Linting, due to the nature of this task.  I can't take responsibility for editing your .vy files to pass the Linter, at least not on this first round.

pandadefi:
Can you fix the files that are falling linting on this PR?

0xbeedao:
> Can you fix the files that are falling linting on this PR? Can you make it so that if registry linting fails it will still do the vault linting and not stop immediately ?

I think so, I'll give it a shot.

contracts/Vault.vy
```diff
@@ -34,13 +36,18 @@
     For further details, please refer to the specification:
     https://github.com/iearn-finance/yearn-vaults/blob/main/SPECIFICATION.md
 """
+from vyper.interfaces import ERC20
 
 API_VERSION: constant(String[28]) = "0.4.3"
 
-from vyper.interfaces import ERC20
-
 implements: ERC20
```
0xbeedao, line 44: I had to move this to the top to get the constants recognized by the linter

contracts/Vault.vy
```diff
@@ -242,8 +271,8 @@ SECS_PER_YEAR: constant(uint256) = 31_556_952  # 365.2425 days
 # `nonces` track `permit` approvals with signature.
 nonces: public(HashMap[address, uint256])
 DOMAIN_SEPARATOR: public(bytes32)
-DOMAIN_TYPE_HASH: constant(bytes32) = keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)')
-PERMIT_TYPE_HASH: constant(bytes32) = keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)")
```
0xbeedao, line 246: I couldn't chop these to 80 characters.

0xbeedao:
That was really a lot of lint, mostly minor stuff.

.gitignore
```diff
@@ -6,3 +6,6 @@ reports/
 node_modules/
 .test_durations
 yarn-error.log 
+flake8.log
+*egg-info/
```
pandadefi, line 10: What is this from?

contracts/Registry.vy
```diff
@@ -99,7 +101,8 @@ def latestRelease() -> String[28]:
     @return The api version of the latest release.
     """
     # NOTE: Throws if there has not been a release yet
-    return Vault(self.releases[self.numReleases - 1]).apiVersion()  # dev: no release
```
pandadefi, line 102: This is breaking the error reporting:
https://eth-brownie.readthedocs.io/en/stable/tests-pytest-intro.html?highlight=%23dev#developer-revert-comments

pandadefi (changes requested):

contracts/Registry.vy
```diff
@@ -99,7 +101,8 @@ def latestRelease() -> String[28]:
...
+    # dev: no release
```
pandadefi, line 104: This is breaking the error reporting:
https://eth-brownie.readthedocs.io/en/stable/tests-pytest-intro.html?highlight=%23dev#developer-revert-comments

.gitignore
0xbeedao: The log is auto-generated by flake8 when it runs
Installing Flake8-vyper-linter via the commandline ended up creating a "flake8...egg-info" file, so I added it to ignore.
That line would be safe to remove, since the egg files are transient.

contracts/Registry.vy
0xbeedao: OK, can roll that back and add a `noqa` to those ones.

pandadefi: Is there a way to improve the linter so that it doesn't change #dev: comments?

0xbeedao:
Pushed an update which should fix those `# dev` changes.

pandadefi:
@0xbeedao the vyper lint tooling will be a bottleneck for future dev since we are upgrading the vyper version when they get released https://github.com/yearn/yearn-vaults/pull/494

We would need to have  `flake8` properly maintained and vyper version agnostic in order to add linting to our CI.

PR #482, opened 2021-10-29
feat: ignore vault airdrops

lbertenasco:


tests/functional/vault/test_shares.py
```diff
@@ -269,10 +269,11 @@ def test_transferFrom(accounts, token, vault):
...
 
 
-def test_do_not_issue_zero_shares(gov, token, vault):
-    token.approve(vault, 500, {"from": gov})
-    vault.deposit(500, {"from": gov})
-    token.transfer(vault, 500)  # inflate price
-    assert vault.pricePerShare() == 2 * 10 ** token.decimals()  # 2:1 price
-    with brownie.reverts():
-        vault.deposit(1, {"from": gov})
+# TODO: fix vault share inflation by using AirdropStrategy
```
lbertenasco, line 272: @fubuloubu @pandadefi I'll need help on this test :)

contracts/Vault.vy
```diff
@@ -1082,7 +1086,6 @@ def withdraw(
             if strategy == ZERO_ADDRESS:
                 break  # We've exhausted the queue
 
-            vault_balance: uint256 = self.token.balanceOf(self)
```
fubuloubu, line 1085: Think you might run into an implicit assumption here that could be causing issues

lbertenasco: sorry i don't follow, what kind of issues?
`vault_balance` is defined above using `self.totalIdle`

contracts/Vault.vy
```diff
@@ -1088,8 +1090,10 @@ def withdraw(
 
     # See @dev note, above.
     value: uint256 = self._shareValue(shares)
+    # memory load to save gas
+    vault_balance: uint256 = self.totalIdle
```
fubuloubu, line 1094: What would happen if there was some exploit that was able to take tokens from the vault?

skeletor-spaceman: I though about this, we can only get exploited on `strategy.withdraw(amount)`.
if we are, it will revert on this step:
```
withdrawn: uint256 = self.token.balanceOf(self) - preBalance
```

skeletor-spaceman: also, the attack would actually find a way to airdrop tokens to the vault on `strategy.withdraw(amount)` to inflate pps price. the way to protect against this is to ask the strategy for the amount of withdrawn tokens to check against our new balance difference

skeletor-spaceman:
Deprecated for v2, moving improvements to v3 design.

PR #484, opened 2021-11-05
feat: decouple vault ownership from the vault

pandadefi:
TODO: 
- [ ] Fix registry
- [ ] Make VaultToken clonable

### DIFF
--- a/contracts/VaultToken.vy
+++ b/contracts/VaultToken.vy
@@ -1,0 +1,186 @@
+# @version 0.2.16
+
+API_VERSION: constant(String[28]) = "0.5.0"
+
+vault: public(address)
+gov: public(address)
+
+event Transfer:
+    sender: indexed(address)
+    receiver: indexed(address)
+    value: uint256
+
+event Approval:
+    owner: indexed(address)
+    spender: indexed(address)
+    value: uint256
+
+decimals: public(uint256)
+balanceOf: public(HashMap[address, uint256])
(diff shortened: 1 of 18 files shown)


contracts/VaultToken.vy
```diff
@@ -0,0 +1,186 @@
...
+    decimals: uint256
+):
+    self.gov = gov
+    self.decimals = decimals
+    # EIP-712
+    self.DOMAIN_SEPARATOR = keccak256(
+        concat(
+            DOMAIN_TYPE_HASH,
+            keccak256(convert("Yearn Vault", Bytes[11])),
+            keccak256(convert(API_VERSION, Bytes[28])),
```
pandadefi, line 47: The domain separator contains the API_VERSION, I am not sure if this should be fetched from the vault since our goal is to update vaults without updating the VaultToken contract. The downside of fetching it from the vault is that if a transaction that includes a signature is broadcated/store while we are updating the vault, the signature won't be valid anymore.

banteg: You can skip it, I think `verifyingContract` and `chainId` fields would provide enough protection.

pandadefi: Are you sure, if we remove that part would it be possible to "replay" the signatures on a newer vault?

pandadefi:
@skeletor-spaceman closing this one, it won't get into this version

PR #485, opened 2021-11-08
feat: split registry

pandadefi:
I would like to split the registry into two contracts, the first keeps the latest vaults for a token `latestVault` the second keeps information about the release version.

This will make it easier to update the vault release/clone method.

Regarding existing data, the first option would be to endorse the latest vaults, second is adding some calls to the VaultRegistry to check data on the existing Registry


storming0x:
Reviewed briefly but you give more context on the problem we are trying to fix and expand on this:

>This will make it easier to update the vault release/clone method.

pandadefi:
My primary goal with this is to have two types of registry one that keeps vault information and one that keeps releases. I think the vault registry should not update and keep getting new data in order to serve it to frontend/subgraph. The release registry on the other hand will be something we can replace from time to time when we need to change the way we clone vaults.

pandadefi:
Another option is to keep the existing registry and use it in collaboration with a ReleaseRegistry.

- When we have a new version:
- A new vault version is added to ReleaseRegistry and Registry.
- The ReleaseRegistry takes care of cloning.

salazarguille:
I reviewed the comments and code.

I agree with creating two registries to differentiate the responsibilities (it makes the code more clear). And more if we will add the logic for cloning vaults.

My concerns are:

- Is the idea to use one of these approaches for v2 or v3 (or both)?
- If the idea is for v2:
   - Regardless of the approach we use, it will impact our subgraph (and I guess Yearn Watch). How big would be the impact? tbh idk, we should analyze it.
   - How will the current vaults/registry live together with the new registries? Will we need to register all the current vaults?
- If the idea is for v3:
   - I think it would be easier because we would be starting from scratch (vaults, addresses, subgraph, etc)

pandadefi:
I would like to start using it now, I see v3 as incremental changes that leads to a new version we can call v3.

> How will the current vaults/registry live together with the new registries? Will we need to register all the current vaults?

We can replay all the events into the new registry or we can have the new registry default to the existing registry for data it wouldn't know about.

Issue #515, opened 2022-01-21
Update token balance to vault share calculation in unit test


gmridul:
In various places at [test_withdrawal.py](https://github.com/yearn/yearn-vaults/blob/main/tests/functional/vault/test_withdrawal.py#L179), the token balance is converted to vault shares via: `token_balance * vault.pricePerShare() // 10 ** vault.decimals()`.

This seems incorrect as we're multiplying `token_balance` with number of tokens each share represents.

The correct calculation should be: `token_balance // vault.pricePerShare() * (10 ** vault.decimals())`.

The test still passes because in all these cases `vault.pricePerShare() // 10 ** vault.decimals()` turns out to be `1`.

PR #516, opened 2022-01-21
fix: update token balance to share calculation

gmridul:
## Description

Fix the way, the number of shares are calculated from token balance. 

In various places at [test_withdrawal.py](https://github.com/yearn/yearn-vaults/blob/main/tests/functional/vault/test_withdrawal.py#L179), the token balance is converted to vault shares via: `token_balance * vault.pricePerShare() // 10 ** vault.decimals()`.

This seems incorrect as we're multiplying `token_balance` with number of tokens each share represents.

The correct calculation should be: `token_balance // vault.pricePerShare() * (10 ** vault.decimals())`.

The test still passes because in all these cases `vault.pricePerShare() // 10 ** vault.decimals()` turns out to be `1`.

Fixes https://github.com/yearn/yearn-vaults/issues/515

## Checklist

- [ ] I have run vyper and solidity linting

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -132,7 +132,7 @@ def increase_pps(CommonHealthCheck, chain):
         vault.setManagementFee(0, {"from": vault.governance()})
         vault.setPerformanceFee(0, {"from": vault.governance()})
         vault.updateStrategyPerformanceFee(strategy, 0, {"from": vault.governance()})
-        strategy.harvest()
+        strategy.harvest({"from": vault.governance()})
         vault.setManagementFee(managementFee, {"from": vault.governance()})
         vault.setPerformanceFee(performanceFee, {"from": vault.governance()})
         chain.sleep(7 * 3600)
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -119,8 +119,8 @@ def strategy(gov, strategist, keeper, rewards, vault, TestStrategy, request):
 
 
 @pytest.fixture
-def increase_pps(CommonHealthCheck, strategy, chain):
-    def increase_pps(vault, token, amount, sender):
+def increase_pps(CommonHealthCheck, chain):
+    def increase_pps(vault, strategy, token, amount, sender):
         common_health_check = vault.healthCheck()
         if common_health_check != ZERO_ADDRESS:
             common_health_check = CommonHealthCheck.at(common_health_check)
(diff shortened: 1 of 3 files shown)


gmridul:
I am not sure how to run solidity and vyper linting. But then, this PR doesn't change and of those files.  I've run python linting via `black`.

gmridul:
> Can you please fix the tests so that price per share isn't 1?

Sure, I'll see how that can be done.

pandadefi:
@gmridul ping

gmridul:
@pandadefi  working on it. Figuring out the easiest way to make the ratio != 1.

pandadefi:
Did you had a look at this:
https://github.com/yearn/yearn-vaults/blob/188af67af7fb7e91d514f0964b3d1acf433345fd/tests/functional/conftest.py#L122

gmridul:
@pandadefi can you take a look now?

tests/functional/conftest.py
```diff
@@ -129,12 +129,12 @@ def increase_pps(vault, token, amount, sender):
         token.transfer(strategy, amount, {"from": sender})
         managementFee = vault.managementFee()
         performanceFee = vault.performanceFee()
-        vault.setManagementFee(0)
-        vault.setPerformanceFee(0)
-        vault.updateStrategyPerformanceFee(strategy, 0)
-        strategy.harvest()
-        vault.setManagementFee(managementFee)
-        vault.setPerformanceFee(performanceFee)
+        vault.setManagementFee(0, {"from": vault.governance()})
```
pandadefi, line 132: You shouldn't change unnecessary parts of the test.

tests/functional/vault/test_shares.py
```diff
@@ -269,10 +269,10 @@ def test_transferFrom(accounts, token, vault):
     assert vault.balanceOf(b) == token.balanceOf(vault)
 
 
-def test_do_not_issue_zero_shares(gov, token, vault, increase_pps):
+def test_do_not_issue_zero_shares(gov, strategy, token, vault, increase_pps):
     token.approve(vault, 500, {"from": gov})
     vault.deposit(500, {"from": gov})
-    strategy = increase_pps(vault, token, 500, gov)  # inflate price
+    strategy = increase_pps(vault, strategy, token, 500, gov)  # inflate price
```
pandadefi, line 275: Why are you changing that?

gmridul: I updated `increase_pps` to have strategy in the argument list. Reason is that `test_progressive_withdrawal()` and `test_withdrawal_with_empty_queue()` use fresh vault and fresh strategies, so the `strategy` supplied by conftest is not the one we want to harvest in `increase_pps()`.

tests/functional/conftest.py
gmridul: After I moved `strategy` in the argument list, when I use it the test, it's reverting because `msg.sender != vault.governance()`, so I explicitly set "from" to `vault.governance()`.

gmridul: If I apply this commit on main branch, it reverts for me: https://github.com/gmridul/yearn-vaults/commit/3b04e62651e4c29ffbbeec61c10385d0973688d5

```
brownie test tests/functional/vault/test_withdrawal.py -k test_withdrawal_with_empty_queue --interactive
```

In any case, I think it's safe to explicitly specify `msg.sender` for these calls.

pandadefi (approved):

tests/functional/conftest.py
gmridul: Thanks for the clean up :)

Issue #478, opened 2021-10-24
Add Slither Support to CI


fubuloubu:
https://github.com/yearn/yearn-vaults/blob/a113064afd59ecac6ae4f8ed9bfe0d0d00c727a3/.github/workflows/lint.yaml#L98

sambacha:
@mohitsaxenaknoldus if you are interested in doing this we can work on something I think will be of better usage, and can provide additional compensation, lmk if you are interested!

mohitsaxenaknoldus:
> @mohitsaxenaknoldus if you are interested in doing this we can work on something I think will be of better usage, and can provide additional compensation, lmk if you are interested!

Yeah sure, just let me know what additional scripts you talking about. I just need more info here on what needs to be done.

PR #520, opened 2022-02-10
chore: add Slither analyzer

fubuloubu:
fixes: #478


sambacha (changes requested):
https://github.com/yearn/yearn-vaults/pull/521/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R150

this will cause this CI process to always fail, as slither will exit with an error if it finds anything, and it will.

see my PR here https://github.com/yearn/yearn-vaults/pull/521 

we can close this thx

arvatran:
It should be possible to ignore the result with a github config, the option here is much simple that what you have made @sambacha

sambacha:
> Seems easier to use `crytic/slither-action@v0.1.0` than your changes @sambacha 
> 
> I will see if we can make crytic/slither-action@v0.1.0` not to fail with a variable

My PR is exactly that 😘

pandadefi (approved):

PR #524, opened 2022-02-15
fix: assert shares amount before withdrawing

rareweasel:
## Description

When withdrawing using a vault's withdraw method the vault will withdraw from strategies in the withdrawal queue if the vault's token balance doesn't directly cover the calculated withdrawal value. Furthermore if depleting the strategies still doesn't cover the necessary withdrawal amount after deducting strategy losses the vault will recalculate the share amount to be burnt.

Fixes # (bug reported in immunefi)

## Checklist

- [x] I have run vyper and solidity linting


pandadefi:
Add some tests please

storming0x (approved):
LGTM!

PR #526, opened 2022-02-20
feat: major refactor of harvest trigger

dudesahn:
## Description

Significant overhaul of the `harvestTrigger` in `baseStrategy`.
- Add check for base fee oracle value. This is a simple contract that checks if the current base fee is above or below a target value, set by the strategist multisig. This allows strategies to prevent keepers from harvesting during gas spikes. 
- Redefine how `minReportDelay` and `maxReportDelay` are handled. `minReportDelay` is now the minimum amount of time we should wait before harvesting, and is gated by the base fee oracle. `maxReportDelay` is the maximum amount of time we should wait, and is not gated by base fee oracle.
- Add `forceHarvestTriggerOnce`. Paired with the baseFee oracle, this allows strategists to signal that a strategy should be harvested the next time that gas prices is favorable. There is an added benefit here of allowing strategists to manually trigger harvests, but without spending ETH (instead spending Yearn's KP3R credits).
- Add `creditThreshold` check. This is a simple check for how much credit a strategy has available from the vault. If it's above the set value, then the trigger is true. This is gated by the base fee oracle.
- Remove `profitFactor` and `debtThreshold`. In practice, these were not used, and caused more problems when they weren't set appropriately high to avoid constantly triggering harvests.
- Added events to emit when calling `forceHarvestTriggerOnce` or when updating `creditThreshold`.

**Update:**
- Added `baseFeeOracle.sol` and updated it to make it more friendly to testing and use on other chains.
- Added ability to set address for the `baseFeeOracle` a strategy uses.
- Added tests for `baseFeeOracle.sol` at maximal coverage.
- 100% coverage on added code for `harvestTrigger`.
- Rebased to latest changes on main `379553f8f0468d62c3f667baa168add99644bb3d`.
- All tests are now passing in development network locally.

## Checklist

- [x] I have run vyper and solidity linting


dudesahn:
All tests in the strategy folder pass. The only failing test locally outside of strategy is tests/functional/vault/test_withdrawal.py::test_token_amount_does_not_change_on_deposit_withdrawal, but this also fails for me on current main branch. It passes, but then immediately errors the same test and freezes.

dudesahn:
cc @pandadefi, for some reason I can't seem to request a review from you

pandadefi:
Your IDE auto format is messing up with a lot of files that shouldn't change

pandadefi:
> All tests in the strategy folder pass. The only failing test locally outside of strategy is tests/functional/vault/test_withdrawal.py::test_token_amount_does_not_change_on_deposit_withdrawal, but this also fails for me locally on current main branch. It passes, but then immediately errors the same test and freezes the testing process and doesn't move further.
> 
> It looks like in the CI, `tests/functional/strategy/test_misc.py::test_harvest_tend_trigger` is failing because I'm pulling the baseFee oracle contract from on-chain in my testing. Not sure if there's a way to get around this, it seems to be network-related since locally I need to specify mainnet-fork for this test to work properly.

Do not use on chain contracts, this will significantly decrease test speed. You will have to find another way around with your test.

pandadefi:
I think we will have to find a way to include this change as an extension of BaseStrategy to keep this repo chain neutral. I suggest you create a new `brownie-strategy-mix` that overrides the BaseStrategy functions with your changes and that can be used to build on top.

flashfish0x:
I don't like the minReportDelay redefining. For one it is confusing as we need to remember it has different meanings for this strategy and for previous ones. And secondly it is useful to stop the strategy uberharvesting if set up wrong. I suggest keeping minReportDelay and adding a new variable for the upper and lower bound of harvesting

flashfish0x:
If you're going to include creditThreshold you should probably also include debtThreshold. I suggest that they are initialised by default to uint(max) as they are problematic and should only be used in advanced situations.

pandadefi:
I guess we can close this one.

PR #531, opened 2022-03-25
chore: fix comment

charlesndalton:


westonnelson (changes requested):
Looks good. 

Perhaps also add: (surplus assets) after the word debt or something similar after the word?

I - and I would imagine others new to Yearn find it a bit confusing the way Yearn uses the term "debt" and didn't know what it was until studying the Vault architecture.

DarkGhost7:
This is the v2 web repo and has be deprecated for v3 https://github.com/yearn/yearn-finance-v3

PR #541, opened 2022-07-15
chore: burn MINIMUM_LIQUIDITY on first deposit

wavey0x:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Checklist

- [ ] I have run vyper and solidity linting
- [ ] I have run the tests on my machine
- [ ] I have followed commitlint guidelines
- [ ] I have rebased my changes to the latest version of the main branch

### DIFF
--- a/contracts/Vault.vy
+++ b/contracts/Vault.vy
@@ -65,7 +65,7 @@ interface HealthCheck:
 
 name: public(String[64])
 symbol: public(String[32])
-decimals: public(uint256)
+decimals: constant(uint256) = 18
 
 balanceOf: public(HashMap[address, uint256])
 allowance: public(HashMap[address, HashMap[address, uint256]])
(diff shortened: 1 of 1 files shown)


contracts/Vault.vy
```diff
@@ -299,6 +300,7 @@ def initialize(
     @param guardian The address authorized for guardian interactions. Defaults to caller.
     """
     assert self.activation == 0  # dev: no devops199
+    assert DetailedERC20(token).decimals() == 18
```
banteg, line 303: this is not the case for all tokens tho? maybe what we really want to do is to force 18 decimal denom for vault shares.

contracts/Vault.vy
```diff
@@ -853,13 +852,17 @@ def _issueSharesForAmount(to: address, amount: uint256) -> uint256:
...
+        self.totalSupply = totalSupply + shares
     else:
-        # No existing shares, so mint 1:1
-        shares = amount
+        # Permanently lock the first MINIMUM_LIQUIDITY tokens
+        toBurn: uint256 = MINIMUM_LIQUIDITY
+        shares = amount - toBurn
+        self.totalSupply = totalSupply + amount
+        self.balanceOf[ZERO_ADDRESS] += toBurn
+        log Transfer(ZERO_ADDRESS, ZERO_ADDRESS, toBurn)
```
banteg, line 862: this would confuse indexers as they won't be able to reconcile the total supply, mistaking this event for a burn.

maybe an easier approach could be to just not allow deposits/withdrawals which result in a total supply in a range of `0 < x < dust`?

storming0x:
should we close this in favor of #542 ?

pandadefi:
closing for https://github.com/yearn/yearn-vaults/pull/542

PR #542, opened 2022-07-20
feat: release 0.4.4

### LINKED ISSUE #153: Upgrade to Solidity 0.8.x Series
Need to see when OpenZeppelin upgrades

pandadefi:
Candidate for next release.
Do not merge until it's ready for a release.

Change Log: 
- Remove balanceOf dependency for estimatedTotalAssets (airdrop protection)
- allow harvest on same block as addStrategy
- upgrade vyper 0.3.3
- upgrade sol 0.8.x
- add deposit/withdraw events
- reduce BaseStrategy size
- add setLockedProfitDegradation event
- add FeeReport event
- add NewPendingGovernance event
- additional testing
- remove BaseWrapper
- remove yToken
- prevent fork EIP712 replay 

closes #153

### DIFF
--- a/contracts/BaseStrategy.sol
+++ b/contracts/BaseStrategy.sol
@@ -878,7 +878,9 @@ abstract contract BaseStrategy {
      */
     function setEmergencyExit() external onlyEmergencyAuthorized {
         emergencyExit = true;
-        vault.revokeStrategy();
+        if (vault.strategies(address(this)).debtRatio != 0) {
+            vault.revokeStrategy();
+        }
 
         emit EmergencyExitEnabled();
     }


contracts/Vault.vy
```diff
@@ -1100,8 +1110,10 @@ def withdraw(
                 continue  # Nothing to withdraw from this Strategy, try the next one
 
             # Force withdraw amount from each Strategy in the order set by governance
+            preBalance: uint256 = self.token.balanceOf(self)
             loss: uint256 = Strategy(strategy).withdraw(amountNeeded)
```
storming0x, line 1114: I was looking for edge cases on loss > amountNeeded and potentially eating the difference between idle and self.token.balanceOf(self), but this math seems to be right

Also remember we have many tests cases around reporting losses and accounting so if we have no regression on those should be good sign

contracts/Vault.vy
```diff
@@ -1736,8 +1753,10 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     #       the Vault based on the Strategy's debt limit (as well as the Vault's).
     totalAvail: uint256 = gain + debtPayment
     if totalAvail < credit:  # credit surplus, give to Strategy
+        self.totalIdle -= credit - totalAvail
         self.erc20_safe_transfer(self.token.address, msg.sender, credit - totalAvail)
     elif totalAvail > credit:  # credit deficit, take from Strategy
+        self.totalIdle += totalAvail - credit
         self.erc20_safe_transferFrom(self.token.address, msg.sender, self, totalAvail - credit)
```
storming0x, line 1760: i think we can add an assumption here in comments that erc20 doesnt have a transfer fee, if thats the case the amount received from strategy may be different than calculated in totalIdle via `totalAvail - credit`

contracts/Vault.vy
```diff
@@ -1793,9 +1812,12 @@ def sweep(token: address, amount: uint256 = MAX_UINT256):
     """
     assert msg.sender == self.governance
     # Can't be used to steal what this Vault is protecting
-    assert token != self.token.address
     value: uint256 = amount
     if value == MAX_UINT256:
         value = ERC20(token).balanceOf(self)
+
+    if token == self.token.address:
+        value = self.token.balanceOf(self) - self.totalIdle
```
storming0x, line 1820: Just to be clear if balanceOF is < totalIdle this is safeMath so it will revert, correct?

Which IMO is ok if this is ever an edge case

tests/functional/vault/test_misc.py
```diff
@@ -124,14 +124,38 @@ def test_negative_available_deposit_limit(Vault, token, gov):
...
 def test_sweep(gov, vault, rando, token, other_token):
-    token.transfer(vault, token.balanceOf(gov), {"from": gov})
-    other_token.transfer(vault, other_token.balanceOf(gov), {"from": gov})
+    # Vault wrapped token reverts on 0 sweep
+    with brownie.reverts():
+        vault.sweep(token, {"from": gov})
 
-    # Vault wrapped token doesn't work
+    # Airdrop want token
+    airdropAmount = token.balanceOf(gov) // 2
```
storming0x, line 132: can we add a similar test outside sweep where we assert an airdrop doesn't change the state of PricePerShare so we have it specd as test

contracts/Vault.vy
```diff
@@ -1736,8 +1753,10 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
     #       the Vault based on the Strategy's debt limit (as well as the Vault's).
     totalAvail: uint256 = gain + debtPayment
     if totalAvail < credit:  # credit surplus, give to Strategy
+        self.totalIdle -= credit - totalAvail
```
storming0x, line 1756: minor gas improvement is to store in memory result of `credit - totalAvail` to use in next line as well

contracts/Vault.vy
```diff
@@ -1736,8 +1753,10 @@ def report(gain: uint256, loss: uint256, _debtPayment: uint256) -> uint256:
...
         self.erc20_safe_transfer(self.token.address, msg.sender, credit - totalAvail)
```
storming0x, line 1757: minor gas improvement is to store in memory result of `totalAvail - credit` to use in next line as well

tests/functional/vault/test_misc.py
```diff
@@ -124,14 +124,38 @@ def test_negative_available_deposit_limit(Vault, token, gov):
...
+    token.transfer(vault, airdropAmount, {"from": gov})
```
storming0x, line 133: we can add a test in deposit test cases for a vault with totalsupply = 0 + airdrop manipulation + first depositor can't benefit from shares to be sure we cover that edge case

storming0x (commented):
LGTM overall, minor comments around a few test cases we should add and gas optimization

contracts/Vault.vy
storming0x: hmm should be memory not storage right? saves an operation i believe @banteg can you confirm?

tests/functional/vault/test_misc.py
storming0x: done in https://github.com/yearn/yearn-vaults/commit/ba735e9b44634cb6da32fafb64fef4d4b1c379cb

contracts/Vault.vy
storming0x: doggo confirm it saves gas but is not worth it for clarity IMO so closing

contracts/Vault.vy
storming0x: same as above

storming0x:
> @storming0x Can you please have a in detail look at the transition from safe math to math, I see one issue was introduced but lucky for us tests got it: [ed04590](https://github.com/yearn/yearn-vaults/commit/ed04590b9b8da187d0b2cfd869bca58954872256)

yeah this one seems to match what we have on the branch here
https://github.com/yearn/yearn-vaults/blob/887d9091b699651843e74f130ba9940fe7053f19/contracts/BaseStrategy.sol#L736

contracts/BaseStrategy.sol
```diff
@@ -751,15 +753,15 @@ abstract contract BaseStrategy {
...
 
         uint256 profit = 0;
-        if (total > params.totalDebt) profit = total.sub(params.totalDebt); // We've earned a profit!
+        if (total > params.totalDebt) profit = (total - params.totalDebt); // We've earned a profit!
 
         // Otherwise, only trigger if it "makes sense" economically (gas cost
         // is <N% of value moved)
         uint256 credit = vault.creditAvailable();
-        return (profitFactor.mul(callCost) < credit.add(profit));
+        return ((profitFactor + callCost) < (credit + profit));
```
invader-tak: `profitFactor * callCost` rather than `(profitFactor + callCost)`

invader-tak: fixed in commit 84994cc4454a677fe4a3f7b159835fc8c0039028

storming0x (approved):
LGTM!

rareweasel (approved):
LGTM, Great job Panda! I left a minor comment.

contracts/BaseStrategy.sol
```diff
@@ -287,46 +285,75 @@ abstract contract BaseStrategy {
...
+        _;
+    }
+
+    modifier onlyVaultManagers() {
+        _onlyVaultManagers();
+        _;
+    }
+
+    function _onlyAuthorized() internal {
+        require(msg.sender == strategist || msg.sender == governance());
```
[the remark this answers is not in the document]
invader-tak: any reason not to swap over to using Custom Errors as you've moved to solc > 8.4?

contracts/Vault.vy
```diff
@@ -1100,8 +1115,10 @@ def withdraw(
                 continue  # Nothing to withdraw from this Strategy, try the next one
 
             # Force withdraw amount from each Strategy in the order set by governance
+            preBalance: uint256 = self.token.balanceOf(self)
             loss: uint256 = Strategy(strategy).withdraw(amountNeeded)
```
wavey0x, line 1119: is it worth changing the withdraw API in BaseStrategy to return `amountFreed` value as well so we don't have to sandwich this function with balance checks?

https://github.com/yearn/yearn-vaults/blob/v0.4.3/contracts/BaseStrategy.sol#L801-L802

invader-tak (approved):
- Recommend replacing `require` with customer errors in a future iteration
 - vyper 3.4 borked the tests for me, but the previous commit using 3.3 looked fine (passed manual inspection/testing)

contracts/Vault.vy
storming0x: That would be a great addition, but i think its out of scope since it breaks backward compatibility, all current strategies wont be able to be ported as is from 0.4.x to this release

storming0x:
> * vyper 3.4 borked the tests for me, but the previous commit using 3.3 looked fine (passed manual inspection/testing)
latest commit rollbacks to 0.3.3 FYI

banteg:
can we squeeze this in? https://github.com/yearn/yearn-vaults/issues/545

contracts/Vault.vy
```diff
@@ -741,7 +763,7 @@ def decreaseAllowance(spender: address, amount: uint256) -> bool:
 
 
 @external
-def permit(owner: address, spender: address, amount: uint256, expiry: uint256, signature: Bytes[65]) -> bool:
+def permit(owner: address, spender: address, amount: uint256, deadline: uint256, v: uint8, r: bytes32, s: bytes32) -> bool:
```
storming0x, line 766: looping in infra

mariuspod:
this is a related issue for the deployed `0.4.4` registry on optimism re camelCased event field names:
https://github.com/pandadefi/registry/issues/9

contracts/Vault.vy
storming0x: agree it makes sense to move to 0.4.5 to avoid confusion , so i guess next steps is release 0.4.4 and then 0.4.5 with this change

storming0x:
> I will merge this into master and cut 0.4.4 release if none express any issues @storming0x @banteg @rareweasel

asking in the protocol group so we are sure that we are all in sync w the plan

contracts/Vault.vy
fubuloubu: As a hack, you can define the `permit` function to be 2612 compliant, and then farm another function name that has the same method ID as `permit(address,address,uint256,uint256,Bytes[65])` and accepts the same type arguments, and then just have that split the signature into components and forward to the correct method

PR #546, opened 2022-10-25
chore: release 0.4.5

pandadefi:


storming0x (approved):
Changes LGTM, needs @dudesahn additional changes before release

contracts/BaseStrategy.sol
```diff
@@ -732,38 +749,33 @@ abstract contract BaseStrategy {
...
+        // Should not trigger if strategy is not active (no assets or no debtRatio)
+        if (!isActive()) return false;
+
+        // check if the base fee gas price is higher than we allow. if it is, block harvests.
+        if (!isBaseFeeAcceptable()) return false;
 
-        // Should not trigger if Strategy is not activated
-        if (params.activation == 0) return false;
+        // trigger if we want to manually harvest, but only if our gas price is acceptable
+        if (forceHarvestTriggerOnce) return true;
```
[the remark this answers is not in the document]
dudesahn: The idea is that by calling `forceHarvestTriggerOnce` you are still respecting the max base fee acceptable—a situation where you want to harvest it but gas is currently too high, you you flip the boolean and know it will harvest later when optimal.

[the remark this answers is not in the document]
dudesahn: If management/gov really needs to harvest immediately, they can either atomically update the baseFeeOracle so that the gas price check will be permissible, or just simply call `harvest()` themselves

dudesahn:
added base fee provider as a submodule linking to a bare bones ape repo, will update deployment addresses there once this is finalized

contracts/BaseStrategy.sol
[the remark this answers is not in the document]
dudesahn: > In which case does the management changes `forceHarvestTriggerOnce` and not call `harvest` themselves?

management would set `forceHarvestTriggerOnce` to `true` when they want keepers to harvest the next time when gas is optimal, instead of doing it now manually themselves when gas may be more expensive. 

additionally, since we use keep3r on mainnet, we don't really "pay" for harvests done by keepers since we already have the LPs bonded, but if we call harvest from the multisig then we have to spend additional ETH on it.

contracts/BaseFeeOracle.sol
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /// @notice Set authorized addresses to set the acceptable base fee
+    function setAuthorized(address _target) external {
+        _onlyGovernance();
+        authorizedAddresses[_target] = true;
+    }
+
+    /// @notice Update our governance address
+    function setGovernance(address _target) external {
```
poolpitako, line 58: no pendingGov logic?

contracts/BaseFeeOracle.sol
```diff
@@ -0,0 +1,82 @@
...
+    function setManualBaseFeeBool(bool _manualBaseFeeBool) external {
+        _onlyAuthorized();
+        manualBaseFeeBool = _manualBaseFeeBool;
+    }
+
+    /// @notice Set authorized addresses to set the acceptable base fee
+    function setAuthorized(address _target) external {
+        _onlyGovernance();
+        authorizedAddresses[_target] = true;
+    }
```
poolpitako, line 55:
```suggestion
    function setAuthorized(address _target, bool _value) external {
        _onlyGovernance();
        authorizedAddresses[_target] = _value;
    }
```

and remove `revokeAuthorized()`

poolpitako (approved):

Schlagonia:
https://github.com/yearn/yearn-vaults/blob/66734ddcd9992392184f68c0f0c288a782ee5c98/contracts/BaseStrategy.sol#L772

You have the comment for liquidateAllPositions above isBaseFeeAcceptable

contracts/BaseStrategy.sol
```diff
@@ -732,38 +749,34 @@ abstract contract BaseStrategy {
...
+        return (vault.creditAvailable() > creditThreshold);
+    }
 
-        // Otherwise, only trigger if it "makes sense" economically (gas cost
-        // is <N% of value moved)
-        uint256 credit = vault.creditAvailable();
-        return ((profitFactor * callCost) < (credit + profit));
+    /**
+     * Liquidate everything and returns the amount that got freed.
+     * This function is used during emergency exit instead of `prepareReturn()` to
```
Schlagonia, line 774: You have the comments for liquidateAllPosition() still here

contracts/BaseStrategy.sol
```diff
@@ -732,38 +749,34 @@ abstract contract BaseStrategy {
...
 
         // Should trigger if hasn't been called in a while
         if ((block.timestamp - params.lastReport) >= maxReportDelay) return true;
 
-        // If some amount is owed, pay it back
-        // NOTE: Since debt is based on deposits, it makes sense to guard against large
-        //       changes to the value from triggering a harvest directly through user
-        //       behavior. This should ensure reasonable resistance to manipulation
-        //       from user-initiated withdrawals as the outstanding debt fluctuates.
-        uint256 outstanding = vault.debtOutstanding();
```
storming0x, line 752: in these versions we no longer use this logic as a trigger ?

contracts/BaseStrategy.sol
```diff
@@ -732,38 +749,34 @@ abstract contract BaseStrategy {
...
-        if (params.activation == 0) return false;
+        // trigger if we want to manually harvest, but only if our gas price is acceptable
+        if (forceHarvestTriggerOnce) return true;
 
         // Should not trigger if we haven't waited long enough since previous harvest
+        StrategyParams memory params = vault.strategies(address(this));
         if ((block.timestamp - params.lastReport) < minReportDelay) return false;
 
         // Should trigger if hasn't been called in a while
         if ((block.timestamp - params.lastReport) >= maxReportDelay) return true;
```
Schlagonia, line 766: Arent these checks back to back redundant? This would only matter if for some reason minReport is > maxReport which doesnt make sense.

contracts/BaseStrategy.sol
dudesahn: yep, no real use for `profitFactor` or `debtThreshold`

contracts/BaseStrategy.sol
dudesahn: ah ty for catching this! updated

contracts/BaseStrategy.sol
dudesahn: it would prevent harvesting the available credit if it hadn't been long enough yet, but in general I do agree that `minReportDelay` isn't super helpful in this setting. 

Maybe worth removing it here and allowing others to implement via `super.harvestTrigger`?

Schlagonia: That would probably be my thought. Seems like most of the time you would want credit available to cause a harvest even when under the minReport unless it was a strategy that has specific lockup times, which should probably get manually implemented in an override version anyway

contracts/BaseFeeOracle.sol
dudesahn: added

contracts/BaseFeeOracle.sol
dudesahn: agree, done

contracts/BaseStrategy.sol
```diff
@@ -489,35 +492,44 @@ abstract contract BaseStrategy {
...
-     *  By default this is 0, meaning any losses would cause a harvest which
-     *  will subsequently report the loss to the Vault for tracking. (See
-     *  `harvestTrigger()` for more details.)
+     *  This may only be called by governance or management.
+     * @param _forceHarvestTriggerOnce Value of true tells keepers to harvest
+     *  our strategy
+     */
+    function setForceHarvestTriggerOnce(bool _forceHarvestTriggerOnce) external onlyVaultManagers {
+        forceHarvestTriggerOnce = _forceHarvestTriggerOnce;
+        emit ForcedHarvestTrigger();
```
rareweasel, line 519: I would add the `bool _forceHarvestTriggerOnce` as parameter, also check we are modifying `forceHarvestTriggerOnce` value [here](https://github.com/yearn/yearn-vaults/pull/546/files#diff-517cca2fbf046cbfefd26edb1fec816e5b2a5571fd99927dc81d6b4be60cd2a7R819) without emitting an event.

rareweasel (changes requested):
Left some minor comments

contracts/BaseStrategy.sol
dudesahn: although we technically can set this to `false` if needed (perhaps, if we accidentally set it to `true` but can't harvest for some reason), the only time practically it will be used is to set to `true` to trigger a harvest, thus I didn't include the bool in the event. does that make sense?

contracts/BaseFeeOracle.sol
```diff
@@ -0,0 +1,114 @@
...
+        _onlyGovernance();
+        authorizedAddresses[_target] = _value;
+    }
+
+    /**
+     * @notice Starts the 1st phase of the governance transfer.
+     * @dev Throws if the caller is not current governance.
+     * @param _governance The next governance address
+     */
+    function setGovernance(address _governance) external {
```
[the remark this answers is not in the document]
dudesahn: Will do. I'll wait on feedback on the below item before pushing the commit.

contracts/BaseStrategy.sol
rareweasel: In general, every time we modify an state variable, we need to emit an event including the new value on it. So, the subgraph can handle the event and get the new value without calling the contract itself directly (it takes more time/resources).

In this specific case, we are modifying `forceHarvestTriggerOnce` (state variable) [here](https://github.com/yearn/yearn-vaults/pull/546/files#diff-517cca2fbf046cbfefd26edb1fec816e5b2a5571fd99927dc81d6b4be60cd2a7R819) without emitting an event, and the new value is not included in the event. So, there is no way for the subgraph to know the current value without calling the contract when the `ForcedHarvestTrigger` is emitted. We might also call the contract to get the value in the Harvest, but these situations are what make the subgraph takes too long to sync (because of all the contract calls it does).

rareweasel (approved):
LGTM, just minor comments

dudesahn:
added indexing for addresses in events as requested :)

PR #548, opened 2022-11-08
wip-partner fee

tommy4241:
- add new role : partner
- add new fee : partnerFee
- add fee distribution mechanism in report func
- added events, functions to set partner & partnerFee
- vault initialiser has been changed accordingly, need to pass partner & partnerFee as params


elie222:
seems to be adding a partner program

